PR #7, opened 2019-09-03
Fetch tags of model and metamodel

jhernand:
Currently the `model` and `metamodel` targets of the _Makefile_ clone
the corresponding repositories only if needed, and then they always do a
`git fetch` to get the latest code. But they don't use the `--tags`
options and as a result some tags may be missing. This patch adds that
`--tags` option.


jhernand:
@nimrodshn Please review.

nimrodshn:
LGTM :100:

PR #10, opened 2019-09-10
Release 0.1.33

jhernand:

- Update to model 0.0.2:
    - Add `DisplayName` attribute to `Subscription` type.

- Update to metamodel 0.0.5:
    - Fix generation of field names for query parameters.
    - Remove `query` and `path` fields from request objects.
    - Remove unused imports.


jhernand:
FYI: @nimrodshn

PR #11, opened 2019-09-10
Check that `make generate` doesn't introduce changes

jhernand:
This patch changes the `pr_check.sh` script so that it checks that
running the `make generate` doesn't introduce changes to the generated
code, as that is a symptom of using the wrong version fo the model or of
the metamodel.


jhernand:
@nimrodshn please review.

nimrodshn:
LGTM :100:

PR #12, opened 2019-09-11
Use about to expire access token if needed

jhernand:
Currently the SDK always tries to refresh the access token when it
expires in less than one minute. But this fails when there is no refresh
token or credentials that can be used to obtain a new one. This patch
changes it so that in that case it uses that access token, even if it is
about to expire.

Related: https://jira.coreos.com/browse/SDA-1402

### DIFF
--- a/token.go
+++ b/token.go
@@ -55,8 +55,7 @@ func (c *Connection) TokensContext(ctx context.Context) (access, refresh string,
 	c.tokenMutex.Lock()
 	defer c.tokenMutex.Unlock()
 
-	// If the access token is expired, then check the refresh token and either refresh it or
-	// request a new one:
+	// Check the expiration times of the tokens:
 	now := time.Now()
 	var accessExpires bool
 	var accessLeft time.Duration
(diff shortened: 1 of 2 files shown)


jhernand:
@zgalor please review.

token.go
```diff
@@ -78,53 +77,120 @@ func (c *Connection) TokensContext(ctx context.Context) (access, refresh string,
 		c.debugExpiry(ctx, "Bearer", c.accessToken, accessExpires, accessLeft)
 		c.debugExpiry(ctx, "Refresh", c.refreshToken, refreshExpires, refreshLeft)
 	}
-	if c.accessToken == nil || (accessExpires && accessLeft < 1*time.Minute) {
-		if c.refreshToken == nil || (refreshExpires && refreshLeft < 1*time.Minute) {
-			c.logger.Debug(ctx, "Requesting new token")
```
zgalor: did you remove this debug log on purpose or by accident?

jhernand: I moved it to the `sendRequestTokenForm` method.

zgalor:
LGTM :+1:

PR #23, opened 2019-10-07
Update to model 0.0.9

jhernand:
following:

- Add `type` attribute to the `ResourceQuota` type.
- Add `config_managed` attribute to the `RoleBinding` type.


jhernand:
FYI: @rawsyntax

PR #31, opened 2019-10-10
Remove `authorizations` directory when generating code

jhernand:
The `auhtorizations` service was recently added to the model, but we
forgot to update the _Makefile_ to remove it when running the `generate`
target. This patch fixes that.


jhernand:
FYI: @tiwillia

PR #32, opened 2019-10-10
Update to ocm-api-model v0.0.12 to add authorizations/v1/[self_]access_review

tiwillia:
Updated `model_version=v0.0.12` in the Makefile, then ran `make generate` to generate the new files.


tiwillia:
```
Juan Hernandez 2:44 PM
Looks good to me. Merge it when CI passes.
```

PR #34, opened 2019-10-10
Add a CONTRIBUTING document detailing release process

tiwillia:
Additionally, this updates the README to mention the authorizations package alongside the accountsmgmt and clustersmgmt packages.

@jhernand PTAL, this should be a good starting point for documentation moving forward. We still need some solid documentation on the expected syntax of the ocm-api-model

I also did not mention ocm-api-metamodel at all, since I have not touched it.


jhernand:
Looks good, thanks @tiwillia.

PR #36, opened 2019-10-10
Update contributing doc to mention examples and link to new ocm-api-model docs

tiwillia:


tiwillia:
@jhernand PTAL and merge if it looks alright.

PR #38, opened 2019-10-12
Don't require POST body

jhernand:
Currently the SDK doesn't accept POST requests without a body. But there
are some resources in the server that required this. For example the
resource that generates access tokens uses the POST method but it
doesn't require any input. This patch removes that restriction from the
SDK.


jhernand:
FYI: @petli-openshift @tiwillia

PR #42, opened 2019-10-24
Update to model 0.0.14

jhernand:
following:

- Added `SKU` type.
- Improved organizations.
- Improved roles.


jhernand:
FYI: @rawsyntax

PR #44, opened 2019-10-24
Update to model 0.0.15

jhernand:
following:

- Added `search` parameter to the accounts `List` method.


jhernand:
FYI: @rawsyntax

PR #45, opened 2019-10-24
Release 0.1.46

jhernand:

- Update to model 0.0.15:
    - Added `search` parameter to the accounts `List` method.


jhernand:
FYI: @rawsyntax

PR #55, opened 2019-11-07
Update to model 0.0.18

jhernand:
following:

- Added support to search role bindings and resource quota.


jhernand:
FYI: @rawsyntax

PR #56, opened 2019-11-07
Release 0.1.51

jhernand:

- Update to model 0.0.18:
** Added support to search role bindings and resource quota.


jhernand:
FYI: @rawsyntax

PR #57, opened 2019-11-08
Update to model 0.0.19

jhernand:
following:

- Added identifiers to role binding type.


jhernand:
FYI: @rawsyntax

PR #58, opened 2019-11-08
Release 0.1.52

jhernand:

- Update to model 0.0.19:
** Added identifiers to role binding type.


jhernand:
FYI: @rawsyntax

PR #61, opened 2019-11-14
Release 0.1.53

jhernand:

- Update to model 0.0.20:
** Query resource quota from root and delete by identifier.

- Update to metamodel 0.0.13:
** Fix imports of `helpers` and `errors` packages.


jhernand:
FYI: @rawsyntax

PR #68, opened 2019-11-19
Update to model 0.0.22

jhernand:
following:

- Add `socket_total_by_node_roles_os` metric query.


jhernand:
FYI: @chambridge

PR #79, opened 2019-11-28
Update to model 0.0.25

jhernand:
following:

- Allow patching role binding.


jhernand:
FYI: @rawsyntax

PR #80, opened 2019-11-28
Release 0.1.62

jhernand:

- Update to model 0.0.25:
** Allow patching role binding.


jhernand:
FYI: @rawsyntax

PR #84, opened 2019-12-04
Update to model 0.0.27

jhernand:
following:

- Add `resource_name` and `resource_cost` attributes to the add-on type.


jhernand:
FYI: @vkareh

PR #85, opened 2019-12-04
Release 0.1.64

jhernand:

- Update to model 0.0.27:
** Add `resource_name` and `resource_cost` attributes to the add-on type.


jhernand:
FYI: @vkareh

PR #86, opened 2019-12-10
Update to model 0.0.28

jhernand:
The more relevant changes in the new versio of the model are the
following:

- Add `AddOnInstallation` type.


jhernand:
FYI: @vkareh

PR #87, opened 2019-12-10
Release 0.1.65

jhernand:

- Update to model 0.0.28:
** Add `AddOnInstallation` type.


jhernand:
FYI: @vkareh

PR #94, opened 2019-12-17
Update to model 0.0.30

jhernand:
The more relevant changes in the new versio of the model are the
following:

- Add support for `ClusterUUID` field.


jhernand:
FYI: @Lir10

PR #95, opened 2019-12-17
Release 0.1.69

jhernand:

- Update to model 0.0.30:
  - Add support for `ClusterUUID` field.


jhernand:
FYI: @Lir10

PR #96, opened 2019-12-17
Add authentication handler

jhernand:
This patch adds an HTTP handler that checks the JWT tokens present in
the `Authorization` header.


jhernand:
[test]

PR #101, opened 2019-12-19
Send 401 instead of 511

jhernand:
The initial version of the authentication handler sends 511 response
codes when there is no `Authentication` header in the request. This is
incorrect because that status code is intended for use by proxy servers.
This patch changes the handler so that it sends 401 instead.


jhernand:
[test]

PR #105, opened 2019-12-19
Update to model 0.0.31

jhernand:
following:

- Add `ban_code` attribute to `Account` type.


jhernand:
FYI: @Lir10

PR #106, opened 2019-12-19
Release 0.1.72

jhernand:

- Update to model 0.0.31:
  - Add `ban_code` attribute to `Account` type.


jhernand:
FYI: @Lir10

PR #115, opened 2020-01-03
Release 0.1.76

jhernand:

- Update to model 0.0.36:
  - Add types and resources for AWS infrastructure access roles.
  - Add GCP flavour and change AWS flavour to contain also the instance type.


jhernand:
FYI: @zgalor @tzvatot

PR #118, opened 2020-01-08
Release 0.1.77

jhernand:

- Don't require Go 1.13.
- Update to model 0.0.37:
** Add new `service_logs` service.
** Add types and resources for machine types.


jhernand:
FYI: @Lir10

PR #124, opened 2020-01-12
Add body details in case of failure to get token

tzvatot:
We see from time to time 400 bad request returning but we do not print the error reason.
This PR should handle it by only catching the 5xx errors at first, and in case of other errors, it will be handled later in the flow.

@jhernand @nimrodshn @vkareh PTAL

### DIFF
--- a/CHANGES.adoc
+++ b/CHANGES.adoc
@@ -3,6 +3,11 @@
 This document describes the relevant changes between releases of the OCM API
 SDK.
 
+== 0.1.79 Jan 9 2020
+
+- Update to metamodel 0.0.22:
+** Fix generation of _OpenAPI_ paths so that all the characters are lower case.
+
 == 0.1.78 Jan 8 2020
 
 - Fix URL prefix for the logs service.
--- a/version.go
+++ b/version.go
@@ -18,4 +18,4 @@ limitations under the License.
 
 package sdk
 
-const Version = "0.1.78"
+const Version = "0.1.79"


nimrodshn:
LGTM :100:

token.go
```diff
@@ -276,7 +276,7 @@ func (c *Connection) sendTokenFormTimed(ctx context.Context, form url.Values) (c
 	// Read the response body:
 	body, err = ioutil.ReadAll(response.Body)
 	if err != nil {
-		err = fmt.Errorf("can't read response: %v", err)
+		err = fmt.Errorf("failed to read response: %v", err)
```
jhernand: This change doesn't belong in this patch.

token.go
```diff
@@ -300,7 +300,7 @@ func (c *Connection) sendTokenFormTimed(ctx context.Context, form url.Values) (c
 	var msg internal.TokenResponse
 	err = json.Unmarshal(body, &msg)
 	if err != nil {
-		err = fmt.Errorf("can't parse JSON response: %v", err)
+		err = fmt.Errorf("failed to parse JSON response: %v", err)
```
jhernand: This change doesn't belong in this patch.

token.go
```diff
@@ -285,7 +285,7 @@ func (c *Connection) sendTokenFormTimed(ctx context.Context, form url.Values) (c
 
 	// Check the response status and content type:
 	code = response.StatusCode
-	if response.StatusCode != http.StatusOK {
+	if response.StatusCode >= 500 {
```
jhernand: I think it would be better to preserve the check for StatusOK and move it to line 314; right after checking if the JSON response contains an error message. Otherwise you are assuming that 5XX responses don't have a JSON body, which may not be true.

PR #126, opened 2020-01-13
Release 0.1.80

jhernand:

- Add body details in case of error from token provider.


jhernand:
FYI: @tzvatot

PR #127, opened 2020-01-14
Add pre request and post request callback support

tzvatot:
@jhernand @nimrodshn @zgalor PTAL


logging_callback.go
```diff
@@ -0,0 +1,42 @@
...
+
+func (c *LoggingCallback) Pre(request *http.Request) {
+	c.logger.Info(request.Context(), "Sending request %s '%s'", request.Method, request.URL.String())
+}
+
+func (c *LoggingCallback) Post(response *http.Response, err error) {
+	if err != nil {
+		c.logger.Error(response.Request.Context(), "Got error sending request")
+	} else {
+		c.logger.Info(response.Request.Context(), "Got response status code %d", response.StatusCode)
```
zgalor: Is it possible that you get an error that is not related to the status code?

send.go
```diff
@@ -31,6 +31,17 @@ import (
 )
 
 func (c *Connection) RoundTrip(request *http.Request) (response *http.Response, err error) {
+	for _, callback := range c.preCallbacks {
+		request = callback.Pre(request)
+	}
+	response, err = c.roundTrip(request)
+	for _, callback := range c.postCallbacks {
+		response, err = callback.Post(response, err)
```
zgalor: Do you have an example to a list of post callbacks?

Theoretically, `err` could be changed inside the callback (depends on implementation), leading the following callback to act upon the previous callback error value.

logging_callback.go
tzvatot: yes, the error can be manipulated by the callback if needed.

send.go
tzvatot: yes, an example would be a logging callback that logs the response status, and a metrics collecting callback to send data for analytics.

jhernand:
I think this should use the `http.RoundTripper` interface instead of adding new interfaces. That interface already allows pre-processing the request and post-processing the response.

tzvatot:
> I think this should use the `http.RoundTripper` interface instead of adding new interfaces. That interface already allows pre-processing the request and post-processing the response.

I think using http.RoundTripper in our exported public SDK is exposing implementation details of the connection.
If we switch to a different http client implementation, or if we upgrade the dependency and the API of http.RoundTripper changes, it will also break the SDK users.
I suggest to hide this implementation details.

jhernand:
The `http.RoundTripper` interface is covered by the Go compatibility promise, which is much stronger than any other thing in the SDK. So using it protects from potential compatibility breaking changs better than the new interfaces that you are proposing.

In addition the `http.RoundTripper` interface is already widely used by pieces of software that could be interesting for users of the SDK. There are already logging round trippers, security round trippers, retry round trippers, etc.

All in all: i understand your concern and anyhow I prefer to use the `http.RoundTripper` interface.

The interface that I would like to see is something like this:

```go
// Create the round tripper:
myRoundTripper := ....

// Create the connection:
connection, err := sdk.NewConnectionBuilder().
        Transport(myRoundTripper).
        Build()
```

I would also like to see tests covering this new feature.

jhernand:
Replaced by #128 .

PR #129, opened 2020-01-15
Fix typos

tzvatot:
@jhernand PTAL

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -302,7 +302,7 @@ func (b *ConnectionBuilder) Insecure(flag bool) *ConnectionBuilder {
 // code, for example if it wasn't possible to open the connection, or if there was a timeout waiting
 // for the response.
 //
-// Note that setting this attribute is not enougth to have metrics published, you also need to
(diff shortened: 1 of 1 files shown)


clustersmgmt/v1/aws_infrastructure_access_roles_client.go
```diff
@@ -103,7 +103,7 @@ func (r *AWSInfrastructureAccessRolesListRequest) Header(name string, value inte
 // The syntax of this parameter is similar to the syntax of the _order by_ clause of
 // a SQL statement, but using the names of the attributes of the role instead of
 // the names of the columns of a table. For example, in order to sort the roles
-// descending by dislay_name the value should be:
```
jhernand: This is generated code. If you change it here the tests will fail because they detect a change in generated code. This typo should be fixed in the model.

The other changes are OK.

tzvatot: Fixed

PR #133, opened 2020-01-21
check non-json response and return error

petli-openshift:
@jhernand @cben pls review.

- check if `content-type` is not `application/json`, raise error so the response is not parsed by clients;
- the error contains some context info so can be logged by client code; (response body is limited to 200 chars)

error example,
```
expected JSON content type but received 'application/html'; request: GET http://127.0.0.1:42469/mypath; response status: 200 200 OK; response body: <textarea class="Playground-input js-playgroundCodeEl" spellcheck="false" aria-label="Try Go">// You can edit this code! // Click here and start typing. package main import "fmt" func main() { fmt.Pri
```

### DIFF
--- a/send.go
+++ b/send.go
@@ -27,6 +27,7 @@ import (
 	"net/http"
 	"path"
 	"strconv"
+	"strings"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


send.go
```diff
@@ -189,5 +190,47 @@ func (c *Connection) send(ctx context.Context, request *http.Request) (response
...
+			response.Header.Get("Content-Type"),
+			getResponseInfo(response))
+		return
+	}
+
+	// check if (maybe) empty
+	if response.Header.Get("Content-Length") == "0" {
+		err = fmt.Errorf("response is empty; %s", getResponseInfo(response))
+		return
+	}
```
jhernand: An empty response body is a perfectly valid response. In particular it is the expected response body for `DELETE`.

petli-openshift: makes sense to me. I removed.

cben:
LGTM :sparkles: 
@jhernand PTAL

PR #135, opened 2020-01-23
Release 0.1.82

jhernand:

- Update to model 0.0.38:
** Add `search` and `order` parameters to the method that lists registry
   credentials.
** Add `labels` parameter to the method that lists subscriptions.
** Add types and resources for management of AWS infrastructure access roles.


jhernand:
FYI: @petli-openshift @zgalor

PR #136, opened 2020-02-03
Update to model 0.0.39

jhernand:
following:

- Add types and resources for cluster operator metrics.
- Add `deleting` and `removed` states to AWS infrastructure access role grant
  status.


jhernand:
[test]

PR #137, opened 2020-02-03
Release 0.1.83

jhernand:

- Check content type of HTTP responses and return an error if it isn't JSON.
- Update to model 0.0.39:
  - Add types and resources for cluster operator metrics.
  - Add `deleting` and `removed` states to AWS infrastructure access role grant
    status.


jhernand:
FYI: @petli-openshift @zgalor @pvasant

PR #141, opened 2020-02-05
Release 0.1.84

jhernand:

- Update to model 0.0.40:
** Add method to update flavour.


jhernand:
FYI: @zvikorn

PR #143, opened 2020-02-12
Update to metamodel 0.0.23

jhernand:
following:

- Fix missing _OpenAPI_ paths due to incorrect use of `append`.


jhernand:
FYI: @pvasant

PR #146, opened 2020-02-13
Release 0.1.86

jhernand:

- Update to model 0.0.41:
  - Add `target_namespace` and `install_mode` attributes to `AddOn` type.
  - Add `state` attribute to `AWSInfrastructureAccessRole` type.


jhernand:
FYI: @zvikorn

PR #151, opened 2020-02-20
Remove service and version parametrs from authentication handler

jhernand:
This patch removes the service and version parameters from the builder
of the authentication handler. These parameters are no longer needed
because now the handler extracts the values from the request URL.

This is a backwards compatibility breaking change that will require
changing the code that creates the authentication handler. For example,
the clusters management service will need to be changed like this:

```diff
diff --git a/cmd/clusters-service/servecmd/apiserver/apiserver.go b/cmd/clusters-service/servecmd/apiserver/apiserver.go
index ee8b40d2..cfaec010 100644
--- a/cmd/clusters-service/servecmd/apiserver/apiserver.go
+++ b/cmd/clusters-service/servecmd/apiserver/apiserver.go
@@ -128,8 +128,6 @@ func (s *Server) Start() {
 		}
 		authnHandler, err := authentication.NewHandler().
 			Logger(authnLogger).
-			Service(api.ID).
-			Version(api.VersionID).
 			Public(fmt.Sprintf("^%s/%s/?$", api.Prefix, api.ID)).
 			Public(fmt.Sprintf("^%s/%s/%s/?$", api.Prefix, api.ID, api.VersionID)).
 			Public(fmt.Sprintf("^%s/%s/%s/openapi/?$", api.Prefix, api.ID, api.VersionID)).
```

And the accounts management service will need to be changed like this:

```diff
diff --git a/pkg/server/server.go b/pkg/server/server.go
index fc027334..9dd7ccf2 100644
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -326,8 +326,6 @@ func NewAPIServer() Server {
 		// Create the handler that verifies that tokens are valid:
 		mainHandler, err = authentication.NewHandler().
 			Logger(authnLogger).
-			Service("accounts_mgmt").
-			Version("v1").
 			KeysFile(env().Config.Server.JwkCertFile).
 			KeysURL(env().Config.Server.JwkCertURL).
 			ACLFile(env().Config.Server.ACLFile).
```

This is specially important for programs that use a single
authentication handler for multiple API services. For example, the
accounts management service uses the same authentication handler for the
`accounts_mgmt` and `authorizations` service, but the authentication
errors generated always use the `/api/accounts_mgmt/v1` prefix.

### DIFF
--- a/authentication/handler.go
+++ b/authentication/handler.go
@@ -46,8 +46,6 @@ import (
 // create objects of this type directly, use the NewHandler function instead.
 type HandlerBuilder struct {
 	logger       sdk.Logger
-	service      string
-	version      string
 	publicPaths  []string
 	keysFiles    []string
 	keysURLs     []string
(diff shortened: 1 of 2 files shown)


jhernand:
FYI: @tiwillia

PR #155, opened 2020-02-25
Support retry when fetching access token

tzvatot:
@jhernand @vkareh @zgalor PTAL

### DIFF
--- a/connection_test.go
+++ b/connection_test.go
@@ -167,27 +167,27 @@ var _ = Describe("Connection", func() {
 		Expect(err).ToNot(HaveOccurred())
 		defer connection.Close()
 		_, _, err = connection.Tokens()
-		Expect(transport.called).To(BeTrue())
+		Expect(transport.called > 2).To(BeTrue()) // it means the retry was called
 		Expect(err).To(HaveOccurred())
 	})
 })
 
 type TestTransport struct {
-	called bool
+	called int
 }
 
 func (t *TestTransport) RoundTrip(request *http.Request) (response *http.Response, err error) {
-	t.called = true
+	t.called++
 	header := http.Header{}
 	header.Add("Content-type", "application/json")
 	response = &http.Response{
 		StatusCode: 401,
 		Header:     header,
-		Body:       gbytes.NewBuffer(),
+		Body:       gbytes.BufferWithBytes([]byte("{}")),
 	}
 	return response, nil
 }
 
 func NewTestTransport() *TestTransport {
-	return &TestTransport{called: false}
+	return &TestTransport{called: 0}
 }
--- a/connection_test.go
+++ b/connection_test.go
@@ -181,7 +181,7 @@ func (t *TestTransport) RoundTrip(request *http.Request) (response *http.Respons
 	header := http.Header{}
 	header.Add("Content-type", "application/json")
 	response = &http.Response{
-		StatusCode: 401,
+		StatusCode: http.StatusInternalServerError,
 		Header:     header,
 		Body:       gbytes.BufferWithBytes([]byte("{}")),
 	}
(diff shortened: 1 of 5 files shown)


send.go
```diff
@@ -115,8 +117,14 @@ func (c *Connection) send(ctx context.Context, request *http.Request) (response
 		return
 	}
 
+	var token string
 	// Get the access token:
-	token, _, err := c.TokensContext(ctx)
+	operation := func() error {
+		token, _, err = c.TokensContext(ctx)
```
vkareh: Not a blocker, but do we want to log retries?

send.go
```diff
@@ -29,6 +29,8 @@ import (
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/cenkalti/backoff"
```
jhernand: According to the documentation of the library this should be `github.com/cenkalti/backoff/v4`.

send.go
```diff
@@ -115,8 +117,14 @@ func (c *Connection) send(ctx context.Context, request *http.Request) (response
 		return
 	}
 
+	var token string
 	// Get the access token:
-	token, _, err := c.TokensContext(ctx)
+	operation := func() error {
+		token, _, err = c.TokensContext(ctx)
+		return err
+	}
```
jhernand: We should retry this only in situations where we know that retry may help. It will not help, for example, if the credentials are incorrect. Can we limit the retry only to situations where the SSO service returns certain error responses?

I'd also suggest to move the retry logic inside the `TokensContext` method, so that users that call directly that method also benefit from that logic.

send.go
tzvatot: Good idea. I added it.

send.go
tzvatot: @jhernand do you know what error responses should trigger a retry?

For the second suggestion - Fixed. Thanks

jhernand:
Currently I believe that this change will not solve the `"invalid_grant": "Session not active"` problem. The reason is that if the session is somehow marked as not active trying again the refresh grant will not solve it. I think what we need to do in that case is discard the current access and refresh tokens and request new ones. That is assuming that this only affects the "client credentials grant", which seems to be the case. I opened #162 to do that and I am closing this.

tzvatot:
@igoihman this was closed for another reason. We may consider bringing this one back to life. WDYT?

igoihman:
@tzvatot yes, I'm reopening it

token.go
```diff
@@ -49,7 +49,30 @@ func (c *Connection) Tokens() (access, refresh string, err error) {
...
+		c.logger.Debug(ctx, "trying to get tokens")
+		var code int
+		code, access, refresh, err = c.tokensContext(ctx)
+		if err != nil {
+			if code >= http.StatusInternalServerError {
+				c.logger.Error(ctx, "failed to get tokens, got http code %d, will attempt to retry. err: %v", code, err)
+				return err
+			}
+			c.logger.Debug(ctx, "failed to get tokens, got http code %d, will not attempt to retry. err: %v", code, err)
+			return nil
```
nimrodshn, line 64: @tzvatot are you dont want to retry even if `err` is non-nil?

nimrodshn:
@tzvatot LGTM apart from my minor comment above ^

token.go
tzvatot: @nimrodshn see @jhernand comment above:
```
We should retry this only in situations where we know that retry may help
```
Currently we suspect that only http 5xx can be recovered by a retry. Not sure about other errors so not retrying for those.

nimrodshn: Alright. :+1:

connection_test.go
```diff
@@ -167,27 +167,27 @@ var _ = Describe("Connection", func() {
 		Expect(err).ToNot(HaveOccurred())
 		defer connection.Close()
 		_, _, err = connection.Tokens()
-		Expect(transport.called).To(BeTrue())
+		Expect(transport.called > 2).To(BeTrue()) // it means the retry was called
```
zgalor, line 170: If retry isn't called, will the test fail?

tzvatot: of course

igoihman:
LGTM

zgalor:
LGTM

zgalor:
LGTM

PR #163, opened 2020-03-02
Release 0.1.90

jhernand:

- Request new tokens when the _OpenID_ server returns error code `invalid_grant`
during the refresh token grant.

- Check that responses from the _OpenID_ server contain `application/json` in
the `Content-Type` header, and improve the error messages generated in that
case so that they contain a summary of the content.

- Honor cookies sent by the _OpenID_ and API servers.

### DIFF
--- a/CHANGES.adoc
+++ b/CHANGES.adoc
@@ -3,6 +3,17 @@
 This document describes the relevant changes between releases of the OCM API
 SDK.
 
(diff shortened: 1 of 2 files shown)


jhernand:
FYI: @Lir10

Lir10:
> FYI: @Lir10

Thanks @jhernand for the update, we will upgrade the version in OSL ASAP

PR #164, opened 2020-03-05
Update to model 0.0.42

jhernand:
following:

- Add `client_secret` attribute to _GitHub_ identity provider.


jhernand:
[test]

PR #170, opened 2020-03-18
Update to model 0.0.46

igoihman:
following:

- Add missing fields for add-on installation
- Add operator name to add-ons


vkareh:
[test]

PR #175, opened 2020-03-22
Update to model 0.0.48

igoihman:
following:

- Fix `OpenID` attributes.
- Add Cluster API listening method.


igoihman:
[test]

igoihman:
@tzvatot @zgalor @vkareh PTAL

clustersmgmt/v1/cluster_api_type.go
```diff
@@ -55,6 +57,29 @@ func (o *ClusterAPI) GetURL() (value string, ok bool) {
...
 
+// Listening returns the value of the 'listening' attribute, or
+// the zero value of the type if the attribute doesn't have a value.
+//
+// The listening method of the API server.
+func (o *ClusterAPI) Listening() ListeningMethod {
+	if o != nil && o.listening != nil {
+		return *o.listening
+	}
+	return ListeningMethod("")
```
zgalor: I am not familiar with the ocm-sdk logic, but is that what we usually do? No default to one of the methods?
If it were a bool, we would probably default to false.

clustersmgmt/v1/listening_method_list_type_json.go
```diff
@@ -26,46 +26,47 @@ import (
 	"github.com/openshift-online/ocm-sdk-go/helpers"
 )
 
-// MarshalOpenIDURLsList writes a list of values of the 'open_IDURLs' type to
+// MarshalListeningMethodList writes a list of values of the 'listening_method' type to
```
zgalor: why is that a list?

clustersmgmt/v1/listening_method_type.go
```diff
@@ -0,0 +1,30 @@
+/*
+Copyright (c) 2019 Red Hat, Inc.
```
zgalor: 2020 :)

clustersmgmt/v1/cluster_api_type.go
igoihman: it's a generated code by `make generate`. The defaults will be handled in clusters-service code

clustersmgmt/v1/listening_method_type.go
igoihman: again, it's generated, the scripts need an update :)

PR #178, opened 2020-03-24
Update to model 0.0.50

vkareh:
following:

- Add Ingress type
- add sockets to cluster_metrics_type


vkareh:
[test]

PR #199, opened 2020-05-29
add a new label to metrics to identify the service

petli-openshift:
unlike api inbound call, these outbound calls can have different destination services. I suggest adding a new label to identify its service,
```
api_outbound_request_duration_sum{apiservice="ocm-clusters-service",code="404",method="GET",path="/api/clusters_mgmt/v1/clusters/-"} 0.791109426
api_outbound_request_duration_count{apiservice="ocm-clusters-service",code="404",method="GET",path="/api/clusters_mgmt/v1/clusters/-"} 1
// from other client...
api_outbound_request_duration_sum{apiservice="quay.io",code="201",method="POST",path="/v2"} 0.791109426
api_outbound_request_duration_count{apiservice="quay.io",code="201",method="POST",path="/v2"} 1
```

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -20,6 +20,7 @@ package sdk
 
 import (
 	"github.com/prometheus/client_golang/prometheus"
+	"strings"
 )
 
 // registerMetrics registers the metrics with the Prometheus library.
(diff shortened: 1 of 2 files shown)


tiwillia:
This makes sense to me. @petli-openshift we should request some review from the SD-A folks also. This will affect the metrics they expose also. Do you know who on that team would be best to review?

send.go
```diff
@@ -67,9 +67,10 @@ func (c *Connection) RoundTrip(request *http.Request) (response *http.Response,
 			code = response.StatusCode
 		}
 		labels := map[string]string{
-			metricsMethodLabel: request.Method,
-			metricsPathLabel:   metric,
-			metricsCodeLabel:   strconv.Itoa(code),
+			metricsAPIServiceLabel: c.GetAPIServiceLabelFromPath(request.URL.Path),
```
cben, line 70: I'm curious, this is different from the way `metricsPathLabel: metric` below obtained the path — that one is taken from X-Metric header.  Will both ways work?

cben:
Aha, so this will contain redundant info with `path` label, and could maybe be simulated at querying time by [label_replace()](https://prometheus.io/docs/prometheus/latest/querying/functions/#label_replace), but yeah why not :+1:

Let me just figure out this weird X-Metric game...  x-men hiclipart com

send.go
petli-openshift: it can be done using the same pattern. It only requires to be set at these very top client builders,
https://github.com/openshift-online/ocm-sdk-go/blob/master/connection.go#L603-L620

However, the issue is we need an extra metric param (or hack the existing one) to it and all its descendant client builders. It requires a lot more code updates. Or there could be better way? pls let me know your thought.

cben:
LGTM :+1:

cben:
@igoihman I guess you can merge here?

petli-openshift:
thanks for looking into this. Currently it's blocking us from adding the same metrics for other clients. It's not urgent, but appreciate if it gets merged by this week.

petli-openshift:
btw. I just added the two missing services: `authentications` and `service-logs`

metrics.go
```diff
@@ -117,11 +118,30 @@ func (c *Connection) registerMetrics(subsystem string) error {
...
+	} else if strings.HasPrefix(path, "/api/authorizations") {
+		return "ocm-authorizations-service"
+	} else if strings.HasPrefix(path, "/api/service_logs") {
+		return "ocm-logs-service"
+	} else {
+		pathParts := strings.Split(path, "/")
+		if len(pathParts) > 3 {
+			pathParts = pathParts[:3]
+		}
+		return "ocm-" + strings.Join(pathParts, "/")
```
vkareh, line 135: Should this join on `"-"` instead of `"/"`?

petli-openshift: I feel this really does not matter. It's just in case it's forgotten when a new service is added. by part of its url, we can tell what goes wrong and fix it accordingly.

PR #210, opened 2020-06-30
Update v0.1.110 changes

tzvatot:
@vkareh @igoihman PTAL


vkareh:
@tzvatot - you need to update `version.go`

tzvatot:
> @tzvatot - you need to update `version.go`

Done

PR #215, opened 2020-07-08
improve error response from gateway

petli-openshift:
the gateway errors are returned with html tags. the goal is to strip these tags and return the real text with the response.

the new error message will be like (see updated test),
```
expected response content type 'application/json' but received 'text/text' and content 'Application is not available The application is currently not serving requests at this endpoint. It may not have been started or is still starting. Possible reasons you are seeing this page: The host ...'
```


Lir10:
LGTM

cben:
Ooh, neat :sparkles:, I'll steal this tag stripping into few more places.  LGTM!

Let's also include http status code in the error (not blocker, can do separately)

send.go
```diff
@@ -198,7 +202,9 @@ func (c *Connection) contentSummary(response *http.Response) (summary string, er
 	if err != nil {
 		return
 	}
-	runes := []rune(string(body))
+	content := strip.StripTags(string(body))
```
cben, line 205: This is done unconditionally but it mostly makes sense for text/html with a long body.
WDYT of checking Content-Type and/or moving this under `if len(runes) > 200`?
Not critical, pretty harmless to do it always, but in that case please update doc comment saying "summary will be the complete body if it isn't too log."

petli-openshift: agreed and updated. will give user raw response unless it's really messy.

petli-openshift: for status code and the status msg, I did have it in the old PR, https://github.com/openshift-online/ocm-sdk-go/pull/133/files#diff-6761b1a148e5dc2caa0297b163135bbcR205-R210

It was removed afterwards. It probably caused some complication. I will leave it to you to decide and add back in the future.

cben:
LGTM :100:

PR #216, opened 2020-07-10
Update sync set example

KevFan:
## Description
Using the previous sync set example gave a `500` error when hitting the sync set api. 

This is due to passing the array of `runtime.RawExtension` as is to the `syncsetBuilder.ID("ext-foo2").Resources()` function. Because this function expects a number of values of type `interface`, the resulting resources body in the json is an array within an array.

Resulting json of the example:
```json
{
  "kind": "Syncset",
  "id": "ext-foo",
  "resources": [
    [
      {
        "kind": "ConfigMap",
        "apiVersion": "v1",
        "metadata": {
          "name": "foo",
          "namespace": "default",
          "creationTimestamp": null
        },
        "data": {
          "foo": "bar"
        }
      }
    ]
  ]
}
``` 

## Changes
* The `resources` array have been updated to be an array of interfaces
* The values from the `resources` array are now passed using the `...` annotation

### DIFF
--- a/examples/create_syncset.go
+++ b/examples/create_syncset.go
@@ -28,7 +28,6 @@ import (
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


zgalor:
/lgtm

Thanks @kevfan

PR #222, opened 2020-08-03
Added support for http PUT method

Lir10:


tiwillia:
Can you please expand methods_test.go to include tests for `PUT`? For the most part, it will be copy/paste of the `PATCH` tests.

vkareh:
Based on offline conversation, we just need to add Put so that we can do authenticated requests from the SDK using the PUT verb. No models support Put at the moment, so no changes are necessary to the model/metamodel repositories.

PR #223, opened 2020-08-03
Release v0.1.116

vkareh:
- Added support for http PUT method
- Update to model v0.0.73
** Add capability_review endpoint
** Add support_cases endpoint


vkareh:
@Lir10 - fyi

PR #224, opened 2020-08-05
bump ocm api model to 0.0.74

tzvatot:
@vkareh @zgalor @igoihman PTAL


vkareh:
@tzvatot - you also need to update `version.go`

PR #240, opened 2020-08-19
token: Remove INFO log when successfully getting token

vkareh:


cben:
This is not enough, there are more info logs, many not interesting to CLI users:

- "OCM auth: got tokens on attempt %d" (#241 fixes it to only be info when retries were needed)
- "OCM auth: server returned error code '%s' and error description '%s' when the refresh token isn't expired"
  (very internal — comment explains this sometimes happens when SSO server restarts or loses its cache)
- "Loading keys from file '%s'"
- "Loading keys from URL '%s'"
- LoggingTransport (opt-in, clusters-service does `TransportWrapper`, not used in ocm nor moactl)
  - "Sending request %s '%s'",
  - "Got response status code %d"

cben:
Copying from private discussion:

> in https://github.com/openshift-online/ocm-sdk-go/pull/231 you have an INFO level log when tokens took more attempts, this shows up in client tools (ocm-cli and moactl) when executing any command, not sure this is a good experience. Is there any particular reason for INFO instead of DEBUG?

> good question, I didn't think of CLI much.
> I do want them in production logs but we don't want to run at debug level.

> is there anything between info and debug? what do prod logs run as?

...
> there is configurable mapping between SDK's logger and actual logger.

> I suspect by default CLI tools don't want any info level logs (unless --debug / --v=1)

> Thought experiment: what if it was OCM API that gave error, and CLI would have a feature to retry (when safe e.g. GET)?
> Do you consider that interesting to user or transparent?

> good question :)  I don't know the answer! haha

...
> fair point - moactl should hide the API from the user, ocm-cli should _maybe_ hide it for user actions (ocm list, create, describe) but show it for http actions (ocm get, post, patch)

cben:
An important point, especially for OCM services that run continually, is there are **no levels between info and debug**.  
At debug level, SDK dumps fulls request & response (censoring some auth fields).  This is not something we normally want.

=> Thus SDK shouldn't move stuff to debug level unless they are really mundane.

This log (for attempt > 1) reflects a problem.  It's was worked around, so it's not error level, but I'd say it's borderline warning, certainly not debug!

The purpose of logging in libaries is precisely "logging".  It's not "stdout humans will see", that's up to the CLI calling the library. A library should be free to add info logs that are useful to "understand what it's doing" without balancing budget constraints of  "black-box use".

WDYT?  I think this means CLI tools should hide info logs by default.

vkareh:
I think this is a reasonable approach.

In this case, I wonder whether we should make the Logger optional when instantiating the client (currently it's required, IIRC). I'll play with the code some more and see how this affects cli tools using the SDK.

Closing for now, thanks for the discussion :)

cben:
I see `ocm` already hides info and even warn, unless --debug or --v=1 is given:
https://github.com/openshift-online/ocm-cli/blob/efb47e13e3c231199801154cbdb1d546a78444aa/pkg/config/config.go#L178-L186

PR #246, opened 2020-08-27
Allow auth header of type AccessToken

rawsyntax:


rawsyntax:
@igoihman @vkareh PTAL

vkareh:
I'm not sure I understand the logic here (I'm not very familiar with the handler interface, to be honest), but shouldn't it call the `next` handler once it confirms that it has the `AccessToken`? So instead of returning, you can just skip to the end of the function (bypassing the bearer/claims check)

authentication/handler_test.go
```diff
@@ -553,6 +553,31 @@ var _ = Describe("Handler", func() {
 		}`))
 	})
 
+	It("Allows access token authorization", func() {
+		// Prepare the next handler, which should not be called:
+		next := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			Expect(true).To(BeFalse())
```
vkareh, line 559: This test will fail now, as the handler is continuing on to `next`

rawsyntax: Fixing up. I was having some issues running the tests locally (passing?)

authentication/handler_test.go
```diff
@@ -553,6 +553,30 @@ var _ = Describe("Handler", func() {
 		}`))
 	})
 
+	It("Allows access token authorization", func() {
+		// Prepare the next handler, which should not be called:
```
vkareh, line 557: `s/should not/should/` :stuck_out_tongue:

rawsyntax: thanks :)

rawsyntax:
@vkareh fixed up

vkareh:
Code /lgtm

Now how do I test it? :)

rawsyntax:
Usage is in a separate repo I've sent you the link

vkareh:
Thanks! /lgtm

PR #249, opened 2020-09-07
update ocm-api-model to 0.0.79

igoihman:


igoihman:
@zgalor @nimrodshn PTAL

PR #257, opened 2020-09-23
add external resources field to add on

ciaranRoche:
JIRA - https://issues.redhat.com/browse/SDA-2805


vkareh:
Let's have a chat to discuss updates to the SDK, but in short we need the ocm-api-model updated first, then the SDK gets autogenerated from the latest version of the model.

Issue #260, opened 2020-09-26
Increase error summaru length


cben:
We've seen errors from SSO/akamai like this:

```Access Denied Access Denied You don\'t have permission to access "http://sso.redhat.com/AK_PM_VPATH0/" on this server. Reference #18.46019b8.16008613... ```

> It looks like the reference numbers in your logs are truncated
> A full reference number looks like Reference #3.2461aad1.1595016997.1e085d88

As these numbers can help SSO team check up on an error, lets increase the truncation limit here:

https://github.com/openshift-online/ocm-sdk-go/blob/a5abe967d77898ddfc1028e061fd0df5daa6ac9b/send.go#L206

PR #261, opened 2020-09-29
increase the limit/size of content summary

ciaranRoche:
FIx Issue -> https://github.com/openshift-online/ocm-sdk-go/issues/260


ciaranRoche:
@igoihman I aired on the side of caution and increased the limit by 50, happy to reduce this to be more exact to the error reference number we are missing. What do you think?

igoihman:
@ciaranRoche thanks, it looks good to me

igoihman (approved):

PR #266, opened 2020-10-05
Bump sdk version

tzvatot:
@igoihman PTAL

### DIFF
--- a/CHANGES.adoc
+++ b/CHANGES.adoc
@@ -3,7 +3,7 @@
 This document describes the relevant changes between releases of the OCM API
 SDK.
 
-== 0.1.134 Oct 5 2020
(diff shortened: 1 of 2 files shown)


vkareh:
what happened with the changelog?

tzvatot:
> what happened with the changelog?

what do you mean?

vkareh:
The change was updating an existing number, not adding a new version, was there an error in the previous bump?

tzvatot:
> The change was updating an existing number, not adding a new version, was there an error in the previous bump?

Yeah, for some unknown reason v.0.1.134 wasn't able to download using go-get. The same procedure worked for 135...

vkareh:
ah got it, thanks

PR #278, opened 2020-10-21
Allow disabling keep alive connections in the SDK connection transport

tiwillia:
I had made this change as part of an investigation into uneven load balancing on the OCM server side. The issue was not keep alives, this change probably has no real purpose and can be closed if desired.

Disabling keep-alive connections significantly increases the amount of time concurrent parallel  requests take to complete.

Note that this is also only disabling HTTP keep-alives, not TCP keep-alives.


vkareh:
I think adding this doesn't hurt

/lgtm

PR #287, opened 2020-11-05
Honour tokens context deadline

jhernand:
Currently the `TokensContext` method uses a back-off strategy to retry failed
requests, but it ignores the deadline of the passed context. This patch fixes
that.

One interesting side effect of that is that it is possible to reduce the time
that tests take to execute by explicitly setting the timeout instead of letting
them run up to the default 15 seconds. This reduces the test time from 45
seconds to 5 seconds.

### DIFF
--- a/token_test.go
+++ b/token_test.go
@@ -19,6 +19,8 @@ limitations under the License.
 package sdk
 
 import (
+	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"strings"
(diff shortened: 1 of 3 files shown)


jhernand:
@tzvatot @tiwillia @cben please review.

token_test.go
```diff
@@ -1056,6 +1058,43 @@ var _ = Describe("Tokens", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+			defer connection.Close()
+
+			// Request the token with a timeout smaller than the artificial
+			// delay introduced by the server:
+			ctx, _ := context.WithTimeout(context.Background(), 5*time.Millisecond)
+			_, _, err = connection.TokensContext(ctx)
+
+			// The request should fail:
+			Expect(err).To(HaveOccurred())
```
tzvatot, line 1096: Can we check the error to make sure it's due to conext deadline reached?

jhernand: Yes, can be checked. It requires changing calls to `fmt.Errorf` to use `%w` instead of `%v` for errors. I am doing that in #289 . Will update this once that is merged.

tiwillia:
LGTM

Issue #319, opened 2021-01-20
Support override of authentication's error handler


JameelB:
Currently, the authentication package has an internal [error handler](https://github.com/openshift-online/ocm-sdk-go/blob/v0.1.152/authentication/handler.go#L779) which cannot be overridden. 

It would be great if we can add support to override this so that we can handle our errors consistently throughout our application which uses this package. Would this be possible?

jhernand:
@JameelB Can you provide an example of the kind of customized error handling that you would like to do?

jhernand:
@JameelB I see. Can you share an example of the JSON response body that you want to send?

JameelB:
@jhernand Here's an example JSON response:

```
{
   "id":"11",
   "kind":"Error",
   "href":"/api/managed-services-api/v1/errors/11",
   "code":"MGD-SERV-API-11",
   "reason":"Bearer token is expired",
   "operation_id":"1nKwdh6ZQOyxSdlsJxOphbGs7Nq"
}
```

jhernand:
Thanks @JameelB. So you need to change the error identifier and the error code, and you need to add an operation identifier for each request. We could provide a generic mechanism to override completely the error reporting, but then it would be an open door to return any kind of thing, including valid responses, and I would prefer to avoid that. Would it be enough for you if we add to the builder of the handler methods to set the error identifier and code, and a method to set a function that returns the operation identifier for the request? Would be something like this:

```go
// Define your own function to extract the operation identifier from
// the request:
func myOperatiIDExtractor(r *http.Request) string {
        return ...
}

// Create the handler using custom error identifer, service identifier
// and operation id extractor:
handler, err := authentication.NewHandler().
        Logger(logger).
        KeysURL(...).
        Service("MGD-SERV-API").
        Error("11").
        OperationID(myOperationIDExtractor).
        Build()
```

Would that satisfy your requirement?

JameelB:
That's perfect for us @jhernand, thank you!

PR #321, opened 2021-01-20
Add support for customizing error responses

jhernand:
This patch adds to the authentication builder the possibility to configure the
error identifier, error code and operation identifier included in JSON error
responses. For example, the following configuration:

```go
handler, err := authentication.NewHandler().
	Logger(logger).
	KeysFile(keysFile).
	Next(next).
	Service("my_service").
	Error("123").
	OperationID(func(r *http.Request) string {
		return r.Header.Get("X-Operation-ID")
	}).
	Build()
if err != nil {
	...
}
```

Will result in JSON error response bodies like this:

```json
{
  "kind": "Error",
  "id": "123",
  "href": "/api/clusters_mgmt/v1/errors/123",
  "code": "MY-SERVICE-123",
  "reason": "Bearer token is malformed",
  "operation_id": "456"
}
```

Note that in this example the operation identifier is taken from the
`X-Operation-ID` request header.

Resolves: https://github.com/openshift-online/ocm-sdk-go/issues/319

### DIFF
--- a/authentication/handler.go
+++ b/authentication/handler.go
@@ -52,6 +52,9 @@ type HandlerBuilder struct {
 	keysCAs      *x509.CertPool
 	keysInsecure bool
 	aclFiles     []string
+	service      string
+	error        string
+	operationID  func(*http.Request) string
 	next         http.Handler
 }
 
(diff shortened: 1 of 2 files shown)


jhernand:
FYI: @JameelB

JameelB:
Thank you @jhernand!

PR #322, opened 2021-01-21
Release 0.1.153

jhernand:

- Add support for customizing the error responses of the authentication handler.
- Add support for connecting to the server using Unix sockets.


jhernand:
@JameelB this version will include the changes that you requested.

PR #326, opened 2021-01-27
Release 0.1.155

jhernand:

- Update to model 0.0.104:
  - Add addon requirement type.


jhernand:
FYI: @mikenairn

PR #329, opened 2021-02-03
connection: Skip loading SystemCertPool on Windows

vkareh:
Since the root system cert pool is not available on Windows we
instantiate an empty cert pool to which we can add the trusted sources.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -27,6 +27,7 @@ import (
 	"net/http/cookiejar"
 	"net/url"
 	"regexp"
+	"runtime"
 	"sort"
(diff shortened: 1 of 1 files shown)


vkareh:
This avoids erroring out due to https://github.com/golang/go/blob/master/src/crypto/x509/cert_pool.go#L106-L109
```go
	if runtime.GOOS == "windows" {
		// Issue 16736, 18609:
		return nil, errors.New("crypto/x509: system root pool is not available on Windows")
	}
```

PR #331, opened 2021-02-04
Accept Empty Reader as non-nil req body

nimrodshn:
Passing a req such as `http.NewRequest('GET', '/api/clusters_mgmt/v1/.., bytes.NewReader(nil))` treated the reader with nil buffer as a non-empty req-body. Should we verify first the length of the body before determining if its non-empty.
cc: @jhernand @igoihman

### DIFF
--- a/send.go
+++ b/send.go
@@ -114,8 +114,8 @@ func (c *Connection) send(ctx context.Context, request *http.Request) (response 
 	switch request.Method {
 	case http.MethodGet, http.MethodDelete:
 		if request.Body != nil {
-			err = fmt.Errorf(
-				"request body is not allowed for the '%s' method",
+			c.logger.Warn(ctx,
+				"Request body is not allowed for the '%s' method",
 				request.Method,
 			)
 			return


send.go
```diff
@@ -114,7 +114,7 @@ func (c *Connection) send(ctx context.Context, request *http.Request) (response
 	switch request.Method {
 	case http.MethodGet, http.MethodDelete:
 		if request.Body != nil {
-			err = fmt.Errorf(
+			c.logger.Warn(ctx,
 				"request body is not allowed for the '%s' method",
 				request.Method,
 			)
```
jhernand, line 120: To be more consistent with other places that send log messages I'd like to have this formatted as follows:

```go
c.logger.Warn(
        ctx,
        "Request body is not allowed for the '%s' method",
        request.Method
)
```

Other than that this looks good to me.

nimrodshn: @jhernand Done.

nimrodshn:
Related issue: https://github.com/tsenart/vegeta/issues/556

PR #333, opened 2021-02-05
Add trusted CA certificates for Windows

jhernand:
This patch adds a new `loadSystemCAs` method to the connection builder, with
two implementations, one for Windows and another one for other operating
systems where Go knows how to load the Trusted CA store. The Windows
implementation explicitly trusts the CAs that we need in order to connect to
the SSO and API services.

This is a temporary hack, till Go learns to load the Windows trusted CA store,
and will break if the SSO or API servers change the CA that they are using.

Related: https://github.com/golang/go/issues/16736
Related: https://github.com/golang/go/issues/18609

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -27,7 +27,6 @@ import (
 	"net/http/cookiejar"
 	"net/url"
 	"regexp"
-	"runtime"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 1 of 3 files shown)


vkareh:
I like this, thanks a bunch!

LGTM

PR #334, opened 2021-02-08
Release v0.1.157

vkareh:
- Accept Empty Reader as non-nil req body
- Missplaced return after warning from send
- Add trusted CA certificates for Windows
- Update metamodel to v0.0.36:
** Use Go 1.15
** Add `documentedSupport` and `namedSupport`
** Add `typedSupport`
** Make reporter streams configurable
** Add presence bitmap
- Update model to v0.0.106:
** Add billing_model field to cluster type
** subscriptions: Add label locator


igoihman:
LGTM

PR #335, opened 2021-02-12
Downgrade to Go 1.14

jhernand:
This patch downgrades the version of the Go required by the SDK from
1.15 to 1.14. The reason to do that is that many users of the SDK are
finding difficulties to migrate to 1.15 due to the lack of that
version in RHEL and to the issues with X.509 certificates and the
obsolete `CN` field that it introduces.

The price to pay is that connections to servers that use TLS over Unix
sockets will be marginally less robust because there is no support Go
1.15 doesn't support passing contexts to the `tls.Dial` function.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/openshift-online/ocm-sdk-go
 
-go 1.15
+go 1.14
 
 require (
 	github.com/cenkalti/backoff/v4 v4.0.0
--- a/send.go
+++ b/send.go
@@ -336,10 +336,17 @@ func (c *Connection) createTransport(ctx context.Context, base *urlInfo) (
 			return dialer.DialContext(ctx, base.network, base.socket)
 		}
 		transport.DialTLSContext = func(ctx context.Context, _, _ string) (net.Conn, error) {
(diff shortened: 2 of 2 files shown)


tiwillia:
LGTM

cben:
Q: is it possible (and is it worth it) to use `// +build !go1.15` vs `// +build go1.15` in the SDK to support 1.14 yet use a more robust implementation when compiled with 1.15+?

jhernand:
@cben Yes, it is possible. I am not sure if it is worth it. I will wait to see someone explicitly requests Go 1.15 support. If so I will use that solution.

The only thing I don't like about the solution itself is that it means that we would silently undo the fix in Go 1.16.

cben:
my understanding (untested) is that version tags are cumulative: each version enables tags for that version _and all previous versions_.
https://golang.org/cmd/go/#hdr-Build_constraints
> - a term for each Go major release, through the current version:
>  "go1.1" from Go version 1.1 onward, "go1.12" from Go 1.12, and so on.

jhernand:
OK, that makes things simpler, thanks @cben.

PR #338, opened 2021-02-15
Update model and release v0.1.159

vkareh:
- Add metrics package
- Update API model to v0.0.107:
** add addon sub operator type


ciaranRoche (commented):
LGTM, thanks @vkareh

PR #340, opened 2021-02-17
Update to model 0.0.108

jhernand:
following:

- Add `billing_model` attribute to the `ReservedResource` type.
- Add `cluster_billing_model` attribute to the `Subscriptioin` type.


jhernand:
FYI: @petli-openshift

PR #360, opened 2021-03-24
errors: Support details []interface{} attribute

vkareh:
Since weberr supports details, OCM errors can contain an array of
arbitrary details. To expose that to consuming clients, we parse it from
the response and add it to the error object.

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -45,6 +45,7 @@ type ErrorBuilder struct {
 	code        string
 	reason      string
 	operationID string
+	details     []interface{}
 }
 
 // Error represents errors.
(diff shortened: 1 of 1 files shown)


jhernand (changes requested):
The file that this is modifying is automatically generated from the model: https://github.com/openshift-online/ocm-api-metamodel/blob/master/pkg/generators/golang/errors_generator.go . If you change it here your change will be lost next time we regenerate the code. Consider changing the metamodel instead.

vkareh:
Ouch! :facepalm: 

I thought it was a stand-alone file, sorry about that!! Will send a PR for the metamodel

PR #365, opened 2021-04-06
Release 0.1.167

jhernand:
The more relevant changes in the new versio are the following:

- Move token logic to separate transport wrapper.
- Use defaults from authentication package.
- Update to model 0.0.113:
** Add RelatedResources struct to QuotaCost.


jhernand:
FYI: @Lir10

PR #376, opened 2021-05-05
Add Fatal logging level

chentex:
Add Fatal logging level

- This will log an Error and execute os.Exit(1) afterwards.
- It is always enabled
- Fix minor errors in checks on std_logger and go_logger

### DIFF
--- a/logging/logger.go
+++ b/logging/logger.go
@@ -30,16 +30,16 @@ import (
 // interface must accept and handle smoothly calls to the Debug, Info, Warn and Error methods where
 // the ctx parameter is nil.
 type Logger interface {
-	// DebugEnabled returns true iff the debug level is enabled.
+	// DebugEnabled returns true if the debug level is enabled.
 	DebugEnabled() bool
 
-	// InfoEnabled returns true iff the information level is enabled.
+	// InfoEnabled returns true if the information level is enabled.
 	InfoEnabled() bool
 
-	// WarnEnabled returns true iff the warning level is enabled.
+	// WarnEnabled returns true if the warning level is enabled.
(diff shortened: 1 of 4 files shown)


logging/logger.go
```diff
@@ -30,16 +30,16 @@ import (
 // interface must accept and handle smoothly calls to the Debug, Info, Warn and Error methods where
 // the ctx parameter is nil.
 type Logger interface {
-	// DebugEnabled returns true iff the debug level is enabled.
+	// DebugEnabled returns true if the debug level is enabled.
```
nimrodshn, line 33: I suspect these where intentional: "if and only if" :shrug:

nimrodshn (commented):
@jhernand  can you please take a look? this looks good to me but I would like to get you're approval as well.

jhernand:
Looks good to me.

The use of "iff" was indeed intentional, but I won't object if you want to remove it. Look like it was confusing to you, may be to others as well.

Note that this is a backwards compatibility breaking change in an interface that will probably be implemented in other projects. Make sure to document it in the `CHANGES.adoc` file and notify people that is using it.

nimrodshn:
@chentex once this is change is done be sure to create a new tagged version of the sdk - To do that please follow this [doc](https://github.com/openshift-online/ocm-sdk-go/blob/master/CONTRIBUTING.adoc#releasing-a-new-ocm-sdk-version).

If you need any help please be sure to contact me or @jhernand.

logging/logger.go
chentex: never crossed my mind, now I know :)

PR #377, opened 2021-05-06
Update `Logger` interface to include `Fatal` level

chentex:
Update `Logger` interface to include `Fatal` level
 ** Fatal level will call `os.Exit(1)` after writing the message.
 ** Fatal level is always active.


jhernand:
Thanks @chentex. Tag v0.1.175 is available now.

PR #378, opened 2021-05-06
Add context to log messages

chentex:
# Why

When using `go_logger` or `std_logger` log messages don't give a context of what type of message is printed

e.g.:
```
ocm-load-test --rate 5/s --duration 1 --test-names self-access-token
2021/05/06 09:55:32 Creating './results' directory
2021/05/06 09:55:32 Using output directory: ./results
2021/05/06 09:55:32 error parsing rate for test self-access-token. Using default
2021/05/06 09:55:32 Executing Test: self-access-token
2021/05/06 09:55:32 Rate: Constant{5 hits/1s}
2021/05/06 09:55:32 Duration: 1m0s
2021/05/06 09:55:32 Endpoint: /api/accounts_mgmt/v1/access_token
2021/05/06 09:56:32 Results written to: 170d1a4b-f9bf-4aaa-853c-f66d67515b1c_self-access-token.json
```

In this case I would expect to see something like this:

e.g.:
```
ocm-load-test --rate 5/s --duration 1 --test-names self-access-token
2021/05/06 09:55:32 Info - Creating './results' directory
2021/05/06 09:55:32 Info - Using output directory: ./results                                                                                                                                                                                         
2021/05/06 09:55:32 Error - error parsing rate for test self-access-token. Using default
2021/05/06 09:55:32 Info - Executing Test: self-access-token
2021/05/06 09:55:32 Debug - Rate: Constant{5 hits/1s}
2021/05/06 09:55:32 Debug - Duration: 1m0s
2021/05/06 09:55:32 Debug - Endpoint: /api/accounts_mgmt/v1/access_token
2021/05/06 09:56:32 Info -Results written to: 170d1a4b-f9bf-4aaa-853c-f66d67515b1c_self-access-token.json
```

If we use `glog_logger` we get by default in their formatting a Prepended character that tells us the type of message we are seeing:

`Lmmdd hh:mm:ss.uuuuuu threadid file:line] msg...` the `L` stands for the log level.

# Changes

So what I tried to do was a general solution that would work with both affected loggers. A new function that will prepend the format for the log message including the Log Level, and it is generic enough that could be extended to make a more complex log format.

Let me know what do you think on this? @nimrodshn @jhernand 

### DIFF
--- a/logging/go_logger.go
+++ b/logging/go_logger.go
@@ -118,6 +118,7 @@ func (l *GoLogger) ErrorEnabled() bool {
 // format and arguments.
 func (l *GoLogger) Debug(ctx context.Context, format string, args ...interface{}) {
 	if l.debugEnabled {
+		format = appendHeader(Debug, format)
 		msg := fmt.Sprintf(format, args...)
 		// #nosec G104
 		log.Output(1, msg)
@@ -128,6 +129,7 @@ func (l *GoLogger) Debug(ctx context.Context, format string, args ...interface{}
 // given format and arguments.
 func (l *GoLogger) Info(ctx context.Context, format string, args ...interface{}) {
 	if l.infoEnabled {
+		format = appendHeader(Info, format)
 		msg := fmt.Sprintf(format, args...)
 		// #nosec G104
 		log.Output(1, msg)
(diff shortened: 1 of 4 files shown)


jhernand:
The logger implementations in this package never intended to be the main loggers used by external applications. They are basically examples of how to write your own adapter for other logging libraries. I prefer to not make these examples more complicated than they are now.

If you want to have the log level as part of the output then I'd suggest that you consider other logging libraries (like logrus or zap) that already include that by default, or can be configured to your linking. Then you can write an adapter for that, using this examples as a guide. That would give you a much better logging infrastructure for your application.

If you still want to use any of these loggers, then you can just copy them to your project and modify them as needed.

chentex:
OK @jhernand makes sense, then I will probably have a wrapper, don't want to have the full logger on my side. I will close this PR.

PR #383, opened 2021-05-18
Update model to v0.0.125 and Release v0.1.178

vkareh:
- Update model to v0.0.125
** Added Custom IAM roles For STS


vkareh:
@pvasant - fyi

PR #396, opened 2021-06-11
Release 0.1.186

vkareh:
- Update model to v0.0.132
** Added Disable workload monitoring
** add authorizations feature review and self feature review models.
** add locators for feature review and self feature review
** Add ccs_only flag for cloud regions


pvasant:
LGTM.. thanks

PR #400, opened 2021-06-23
Don't require refresh token for client credentials grant

jhernand:
Currently the authentication transport wrapper requires the presence of
the refresh token in all responses from the SSO server. This used to be
true in older versions of the server behind `sso.redhat.com`, but
recently it was changed to not return it for client credentials grant
requests. That means that the SDK will fail to obtain a new token when
the initial one is expired. This patch fixes that issue.

Related: https://issues.redhat.com/browse/SDA-4352


cben:
Logic changes all LGTM :clap: 

Ideally during the transition we'd have _more_ test cases — both for old behavior and upcoming new behavior we want to be compatible with.
I see there were a few test cases you removed, which won't make sense with new SSO behavior, but since our accounts are still special-cased to see the old SSO behavior, perhaps they're worth keeping (converted to `RespondWithAccessAndRefreshTokens`)?

PR #410, opened 2021-07-06
Update to metamodel 0.0.38

jhernand:
The only change in the new version of the metamodel is that the
generated code explicitly imports `github.com/golang/glog` in order to
avoid conflicts with `github.com/istio/glog`.


jhernand:
FYI: @zgalor @tzvatot

PR #411, opened 2021-07-06
Release 0.1.194

jhernand:

- Update to version 0.0.38 of the metamodel in order to fix a conflict
  between `github.com/golang/glog` and `github.com/istio/glog` that
  prevents building packages that use the SDK.


jhernand:
FYI: @zgalor @tzvatot

PR #413, opened 2021-07-07
Add auth pkg from clusters-service

nimrodshn:
cc @tzvatot  @jhernand  we also have `ocm-common` which might be suited for this code. wdyt?

### DIFF
--- a/auth/auth_selector_handler.go
+++ b/auth/auth_selector_handler.go
@@ -1,0 +1,36 @@
+package auth
+
+import (
+	"net/http"
+)
+
+type SelectFunc = func(r *http.Request) bool
+
+type HandlerSelection struct {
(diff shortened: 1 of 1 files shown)


jhernand:
I'd say that this can be added to the SDK, but before doing so there are a few thing I will ask:

1. This handler doesn't seem to do anything related to authentication, it looks completely generic. If that is intended then it shouldn't be in the `auth` packge, but probably in a new `http` package.

2. If it does something related to authentication then it should go in the already existing `authentication` package.

3. It should follow the conventions used by other types in the SDK: should have a builder with methods to configure it.

4. Public methods should be documented.

5. It should have tests.

6. If possible an example showing how to use it should be added to the `examples` directory.

PR #415, opened 2021-07-11
Add `transaction` package with logic for transaction handling

nimrodshn:
cc @jhernand @tzvatot

### DIFF
--- a/transaction/context.go
+++ b/transaction/context.go
@@ -1,0 +1,42 @@
+/*
+Copyright (c) 2021 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package transaction
+
+import (
+	"context"
+	"fmt"
+)
+
+type contextKey int
+
+const (
+	transactionKey contextKey = iota
+)
+
+// IntoContext returns a new context with the transaction unit stored in it.
+func IntoContext(u *Unit, ctx context.Context) (context.Context, error) {
+	return context.WithValue(ctx, transactionKey, u), nil
+}
+
+// FromContext Retrieves the transaction from the context.
+func FromContext(ctx context.Context) (*Unit, error) {
+	transactionUnit, ok := ctx.Value(transactionKey).(*Unit)
+	if !ok {
+		return nil, fmt.Errorf("Could not retrieve transaction from context")
+	}
+	return transactionUnit, nil
+}
--- a/transaction/transaction.go
+++ b/transaction/transaction.go
@@ -1,0 +1,210 @@
+/*
+Copyright (c) 2021 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package transaction
+
+import (
+	"context"
+	"database/sql"
+	"runtime/debug"
+
+	"github.com/openshift-online/ocm-sdk-go/logging"
+)
+
+// By default do no roll back transaction only if it was set explicitly using rollback.Flag(ctx).
+const defaultRollbackPolicy = false
+
+// ManagerBuilder contains the data and logic needed to create
+// a transaction manager.
+type ManagerBuilder struct {
+	logger logging.Logger
+	db     *sql.DB
+}
+
+// Manager knows how to start, resolve and in general manage database
+// transactions.
+type Manager struct {
+	logger logging.Logger
+	db     *sql.DB
+}
+
+// Unit represents an sql transaction unit
+type Unit struct {
+	// Reference to the database transaction:
+	tx *sql.Tx
+
+	// Flag indicating if the transaction should be rolled back:
+	rollbackFlag bool
+
+	// List of functions to call after committing the transaction:
+	postCommitCallbacks []func()
+
+	// List of functions to call after rolling back the transaction:
+	postRollbackCallbacks []func()
+}
+
+// Resolve resolves the current transaction according to the rollback flag.
+func (u *Unit) Resolve() error {
+	if u.rollbackFlag {
+		if err := u.rollback(); err != nil {
+			return err
+		}
+	} else {
+		if err := u.commit(); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// commit commits the transaction stored in context or returns an err if one occurred.
+func (u *Unit) commit() error {
+	// Commit the transaction:
+	err := u.tx.Commit()
+	if err != nil {
+		return err
+	}
+
+	// Run the post commit callbacks:
+	for _, callback := range u.postCommitCallbacks {
+		if callback != nil {
+			go callback()
+		}
+	}
+
+	return nil
+}
+
+// rollback rollbacks the transaction stored in context or returns an err if one occurred..
+func (u *Unit) rollback() error {
+	// Rollback the transaction:
+	err := u.tx.Rollback()
+	if err != nil {
+		return err
+	}
+
+	// Run the post rollback callbacks:
+	for _, callback := range u.postRollbackCallbacks {
+		if callback != nil {
+			go callback()
+		}
+	}
+
+	return nil
+}
+
+// MarkForRollback flags the transaction stored in the context for rollback.
+func (u *Unit) MarkForRollback() {
+	u.rollbackFlag = true
+}
+
+// AddPostCommitCallback adds a callback function that will be executed after the transaction is
+// successfully committed. If multiple callbacks functions are added then the order of their
+// execution isn't guaranteed, and they may run in parallel in different goroutines.
+func (u *Unit) AddPostCommitCallback(callback func()) error {
+	u.postCommitCallbacks = append(u.postCommitCallbacks, callback)
+	return nil
+}
+
+// AddPostRollbackCallback adds a callback function that will be executed after the transaction is
+// rolled back. If multiple callbacks functions are added then the order of their
+// execution isn't guaranteed, and they may run in parallel in different goroutines.
+func (u *Unit) AddPostRollbackCallback(callback func()) error {
+	u.postRollbackCallbacks = append(u.postRollbackCallbacks, callback)
+	return nil
+}
+
+// NewManager creates a builder that can then be used to configure and
+// create a new transaction manager.
+func NewManager() *ManagerBuilder {
+	return &ManagerBuilder{}
+}
+
+// Logger sets the object that the manager will use to write messages to the log.
+func (b *ManagerBuilder) Logger(value logging.Logger) *ManagerBuilder {
+	b.logger = value
+	return b
+}
+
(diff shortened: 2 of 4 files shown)


http/transactions_middleware.go
```diff
@@ -0,0 +1,110 @@
+// This file contains an HTTP middleware that associates a transaction to each request.
```
jhernand, line 1: Missing copyright header.

transaction/README.md
```diff
@@ -0,0 +1,41 @@
+# SQL Transaction Library:
```
jhernand, line 1: The documentation format for this project is AsciiDoc.

http/transactions_middleware_test.go
```diff
@@ -0,0 +1,201 @@
+package http
```
jhernand, line 1: Missing copyright header.

transaction/context.go
```diff
@@ -0,0 +1,180 @@
+package transaction
```
jhernand, line 1: Missing copyright header.

transaction/dummy_transaction_service.go
```diff
@@ -0,0 +1,20 @@
+package transaction
```
jhernand, line 1: MIssing copyright header.

transaction/service.go
```diff
@@ -0,0 +1,11 @@
+package transaction
```
jhernand, line 1: MIssing copyright header.

transaction/transaction.go
```diff
@@ -0,0 +1,65 @@
+package transaction
```
jhernand, line 1: Missing copyright header.

transaction/transaction.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"database/sql"
+)
+
+// db is the global database connection object.
+var db *sql.DB
```
jhernand, line 8: I would like to avoid use of global varaiables where possible. Instead of having this global variable I'd suggest to have a new `Manager` type, something like this:

```go
// ManagerBuilder contains the data and logic needed to create
// a transaction manager.
type ManagerBuilder struct {
        logger logging.Logger
        db *sql.DB
        ...
}

// Manager knows how to start, resolve and in general manage database
// transactions.
type Manager struct {
        logger logging.Logger
        db *sql.DB
        ...
}

// NewManager creates a builder that can then be used to configure and
// create a new transaction manager.
func NewManager() *Manager {
        return &ManagerBuilder{}
}

// Logger sets the object that the manager will use to write messages to the log.
func (b *ManagerBuilder) Logger(value logging.Logger) *ManagerBuilder {
        b.logger = value
        return b
}

// Handle sets the database handle that the manager will use to create transactions.
func (b *ManagerBuilder) Handle(value *sql.DB) *ManagerBuilder {
        b.handle = value
        return b
}

// Build uses the data stored in the builder to create a new transaction manager.
func (b *ManagerBuilder) Build(ctx context.Context) (result *Manager, err error) {
        ...
}
```

http/transactions_middleware.go
```diff
@@ -0,0 +1,110 @@
...
+package http
```
jhernand, line 2: I think this should go in the `transaction` package.

http/transactions_middleware.go
```diff
@@ -0,0 +1,110 @@
...
+
+import (
+	"encoding/json"
+	"net/http"
+
+	"gitlab.cee.redhat.com/service/uhc-clusters-service/pkg/global"
+	"gitlab.cee.redhat.com/service/uhc-clusters-service/pkg/transaction"
```
jhernand, line 9: This package shouldn't import anything from the `uhc-clusters-service` module.

http/transactions_middleware.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+// TransactionResponseWriter is a wrapper for http.ResponseWriter,
+// it attempts to resolve (commit or rollback) the transaction on the context of the request
+// if the transaction failed for some reason it catches the error and returns a 500 response
+// with a clear error indicating the failure.
+type TransactionResponseWriter struct {
```
jhernand, line 16: This is going to be used only as in implementation detail of the middleware, so it should be private.

http/transactions_middleware.go
```diff
@@ -0,0 +1,110 @@
...
+func (t *TransactionResponseWriter) WriteErrIfNothingWritten(r *http.Request) {
+	if !t.headerWritten {
+		body := E500.Format(r, "")
+		SendError(t.ResponseWriter, r, &body)
+	}
+}
+
+// TransactionMiddleware creates a new HTTP middleware that begins a database transaction and stores
+// it in the request context.
+func TransactionMiddleware(next http.Handler) http.Handler {
```
jhernand, line 80: This "middleware" should use the same style used by other HTTP handler wrappers in this package. It should be named `HandlerWrapper` have a builder to configure it. I'd expect to use it like this:

```go
// Create the transaction manager:
manager, err := transaction.NewManager().
        Logger(logger).
        Handle(handle).
        Build(ctx)
if err != nil {
        ...
}

// Create the transaction handler wrapper:
wrapper, err := transaction.NewHandlerWrapper().
        Logger(logger).
        Manager(manager).
        Build(ctx)
if err != nil {
        ...
}

// Wrap the handler:
wrappedHandler := wrapper.Wrap(originalHandler)
```

http/transactions_middleware_test.go
```diff
@@ -0,0 +1,201 @@
...
+	"net/http"
+	"net/http/httptest"
+	"strings"
+
+	// nolint
+	. "github.com/onsi/ginkgo"
+	// nolint
+	. "github.com/onsi/gomega"
+
+	sqlmock "github.com/DATA-DOG/go-sqlmock"
```
jhernand, line 16: Instead of using a mock I'd prefer to run the tests using a real database. There is some infrastructure for that in the `testing` package. See the `leadership/flag_test.go` file for examples.

transaction/context.go
```diff
@@ -0,0 +1,180 @@
...
+	tx, err := new()
+	if err != nil {
+		return nil, err
+	}
+
+	return context.WithValue(ctx, transactionKey, tx), nil
+}
+
+// TxContext creates a new transaction context from context.Background()
+func TxContext() (ctx context.Context, err error) {
```
jhernand, line 30: I think we should have a public `Transaction` type. This function should not create that transaction, it should only put it in the context. It should then be named `IntoContext` and shouldn't return an error. For example:

```go
// Create the unit of work:
var transaction.Unit unit = manager.Begin(ctx)

// Put it into the context:
ctx = transaction.IntoContext(ctx, unit)
```

transaction/context.go
```diff
@@ -0,0 +1,180 @@
...
+	return NewContext(context.Background())
+}
+
+// FromContext Retrieves the transaction from the context.
+func FromContext(ctx context.Context) (*sql.Tx, error) {
```
jhernand, line 35: This should return the unit of work and no error:

```go
// Get the current unit of work:
unit := transaction.FromContext(ctx)
```

transaction/context.go
```diff
@@ -0,0 +1,180 @@
...
+func getTransaction(ctx context.Context) (*transaction, error) {
+	transaction, ok := ctx.Value(transactionKey).(*transaction)
+	if !ok {
+		return nil, errors.Errorf("Could not retrieve transaction from context")
+	}
+	return transaction, nil
+}
+
+// Resolve resolves the current transaction according to the rollback flag.
+func Resolve(ctx context.Context) error {
```
jhernand, line 52: `Resolve` should be a method of the `Unit` type.

transaction/context.go
```diff
@@ -0,0 +1,180 @@
...
+		if err := commit(ctx); err != nil {
+			global.Logger.Error(ctx, "Could not commit transaction: %v", err)
+			return err
+		}
+	}
+	return nil
+}
+
+// MarkForRollback flags the transaction stored in the context for rollback.
+func MarkForRollback(ctx context.Context) {
```
jhernand, line 75: `MarkForRollback` should be a method of the `Unit` type.

transaction/context.go
```diff
@@ -0,0 +1,180 @@
...
+		return
+	}
+	global.Logger.Debug(ctx, "Marked transaction for rollback.")
+	transaction.rollback = true
+}
+
+// AddPostCommitCallback adds a callback function that will be executed after the transaction is
+// successfully committed. If multiple callbacks functions are added then the order of their
+// execution isn't guaranteed, and they may run in parallel in different goroutines.
+func AddPostCommitCallback(ctx context.Context, callback func()) error {
```
jhernand, line 88: `AddPostCommitCallback` should be a method of the `Unit` type.

transaction/context.go
```diff
@@ -0,0 +1,180 @@
...
+		return err
+	}
+	transaction.postCommitCallbacks = append(transaction.postCommitCallbacks, callback)
+	return nil
+}
+
+// AddPostRollbackCallback adds a callback function that will be executed after the transaction is
+// rolled back. If multiple callbacks functions are added then the order of their
+// execution isn't guaranteed, and they may run in parallel in different goroutines.
+func AddPostRollbackCallback(ctx context.Context, callback func()) error {
```
jhernand, line 100: `AddPostRollbackCallback` should be a method of the `Unit` type.

transaction/context.go
```diff
@@ -0,0 +1,180 @@
...
+	transaction, err := getTransaction(ctx)
+	if err != nil {
+		return err
+	}
+	transaction.postRollbackCallbacks = append(transaction.postRollbackCallbacks, callback)
+	return nil
+}
+
+// RecoverPanic recovers from a panic if one such occurred.
+func RecoverPanic(ctx context.Context, failure interface{}) {
```
jhernand, line 110: This should be a method of the `Manager` type. And it should not swallow the panic; it should instead re-panic so that other components can also process that panic should they want.

jhernand (changes requested):

transaction/context.go
nimrodshn: The issue is that this should be called inside a `recover` block - that is so if were panicking _after_ a `recover` I dont see the user for that semantics? 

Heres an example inside the middleware:
```
func TransactionMiddleware(next http.Handler, manager *Manager) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		// Create a new Context with the transaction stored in it.
		u, err := manager.Begin()
		if err != nil {
			errors.SendInternalServerError(w, r)
		}
		ctx, err := IntoContext(u, r.Context())
		if err != nil {
			errors.SendInternalServerError(w, r)
			return
		}

		// Make a copy of the request context with the new context key,value stored in it.
		reqWithTx := r.WithContext(ctx)

		txWriter := TransactionResponseWriter{
			ResponseWriter: w,
			request:        reqWithTx,
			headerWritten:  false,
		}

		defer func() {
			if f := recover(); f != nil {
				manager.RecoverPanic(reqWithTx.Context(), f) // Here weve already panicked.
				txWriter.WriteErrIfNothingWritten(r)
			}
		}()

		// Continue handling requests.
		next.ServeHTTP(&txWriter, reqWithTx)
	})
}
```

nimrodshn:
@jhernand Please take a look - we might need to change the Unit tests.

transaction/transaction.go
```diff
@@ -0,0 +1,210 @@
...
+		// Rollback transaction to avoid committing partial writes.
+		u.MarkForRollback()
+		err = u.Resolve()
+		if err != nil {
+			if err == sql.ErrTxDone {
+				return
+			}
+			return
+		}
+		panic(failure)
```
nimrodshn, line 184: @jhernand Re-panicking here ^^

transaction/transactions_middleware.go
```diff
@@ -0,0 +1,122 @@
...
+func (t *TransactionResponseWriter) Write(payload []byte) (int, error) {
+	// Some applications may start writing the response body before calling the `WriteHeader`
+	// method. That is completely OK according to the documentation. But if we don't call it
+	// explicitly the transaction will never be resolved and the connection will be leaked.
+	if !t.headerWritten {
+		t.WriteHeader(http.StatusOK)
+	}
+	if t.transactionErr != nil {
+		errors.SendInternalServerError(t.ResponseWriter, t.request)
+		return
```
nimrodshn, line 77: I tried to align with other `ocm-sdk-go` APIs using `SendInternalServerError` however using it this way I'm not sure if the length of the body written to the `ResponseWriter` by `SendInternalServerError` is reflected in the response and isn't just returning zero bytes.

Issue #421, opened 2021-08-06
request: Making ocm-sdk easily packageable


Gundersanne:
https://github.com/osbuild/osbuild-composer/ is a project that's aiming to run on `api.openshift.com` in future. However it also  releases regularly into fedora/epel etc... And in order to depend on the ocm-sdk a few minor changes could help us greatly.

Most of ocm-sdk's dependencies are release into these platforms already, except:

- github.com/grokify/html-strip-tags-go:
  alternative https://github.com/microcosm-cc/bluemonday is packaged, plus it's much better suited to handle untrusted html.
- https://gitlab.com/c0b/go-ordered-json:
  a bit of a weird one, doesn't have go module support and seems really unmaintained (last code update was 3 years ago).
  We could package it alongside, but it has bugs (the tests don't work). Could also just copy the needed file into this project with a reference to it's upstream.

---

There's also the issue that https://github.com/dgrijalva/jwt-go is unmaintained. https://github.com/golang-jwt/jwt is a maintained fork, which sadly isn't yet packaged, [but hopefully will be soon](https://bugzilla.redhat.com/bugzilla/show_bug.cgi?id=1989386).

---

I'm happy to do the work on this, but want to discuss if it would be acceptable from your side first.

vkareh:
Another option is to vendor in the dependencies instead of relying on packaged versions of those modules. That way the release cycle of ocm-sdk-go is not tied to the release cycle of third-party modules.

Gundersanne:
Yea that's what composer ends up doing for most  distributions other than fedora, though vendoring has it's own drawbacks. composer is tied to release cycles of distributions and thus the packages available on those distros, and I guess I'm trying to reconcile it being both tied to distribution's release cycles and being able to run it on a.o.c.

That said we'd package ocm-sdk, so I wouldn't be asking for openshift/ocm to actually package/release it into fedora.

Regardless some of the issues mentioned in the ticket re:deprecation are still valid I think. But feel free to close this ticket if that's best.

jhernand:
@Gundersanne I think this is a very reasonable request. Some of the changes we should do regardless of packaging. I am reopening and starting working on it.

Gundersanne:
Awesome, if you want me to take on one of them, like the tag stripping one, I'm very happy to do so.

jhernand:
@Gundersanne yes, please, take the tag stripping. I will start with migrating to the new JWT library.

Gundersanne:
[...]

PR #422, opened 2021-08-09
internal: Switch to bluemonday to strip tags from html content

Gundersanne:
As discussed in #421, might need a better commit message though.


jhernand (approved):
Looks good, thanks @Gundersanne.

PR #424, opened 2021-08-10
Use `golang-jwt/jwt` instead of `dgrijalva/jwt-go`

jhernand:
The `dgrijalva/jwt-go` library is no longer maintained and `golang-jwt/jwt` is a
community maintained fork. See https://github.com/dgrijalva/jwt-go/issues/462
for detailts.

Parts of the public interface of the SDK use this library, so this is a
backwards compatibility breaking change. Projects using the SDK will need to
switch to the new library, specially if they are using the
`context.ContextWithToken` or `context.TokenFromContext` functions. The change
should only require changing the import paths, as the fork is fully compatible
with the original library.

A simple way to do the required changes is the following command:

```
$ find . -name '*.go' | xargs sed -i 's|dgrijalva/jwt-go|golang-jwt/jwt|'
```

This also addresses CVE-2020-26160, but that vulnerability doesn't
currently affect the SDK because the authentication handler doesn't use
the `aud` claim.

Related: https://github.com/openshift-online/ocm-sdk-go/issues/421
Related: https://github.com/dgrijalva/jwt-go/issues/462
Related: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-26160

### DIFF
--- a/CHANGES.adoc
+++ b/CHANGES.adoc
@@ -3,6 +3,32 @@
 This document describes the relevant changes between releases of the OCM API
 SDK.
 
+== 0.1.199 TBD
+
+- Use `golang-jwt/jwt` instead of `dgrijalva/jwt-go`.
++
+The `dgrijalva/jwt-go` library is no longer maintained and `golang-jwt/jwt` is
+a community maintained fork. See https://github.com/dgrijalva/jwt-go/issues/462
+for detailts.
++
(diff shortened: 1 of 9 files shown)


jhernand:
@Gundersanne please review.

Gundersanne (approved):
Nice, thank you! Grepped through and can't find any other references, nice that it's just a drop-in replacement.

PR #426, opened 2021-08-10
Fix minor JSON debug issue

jhernand:
We need to convert the buffer to string before printing, otherwise the
output looks like this:

```
2021/08/10 13:33:53 {{
  "kind": "ClusterList",
  "page": 1,
  "size": 2,
  "total": 2,
  "items": [
    ...
  ]
} %!s(int=0) %!s(bytes.readOp=0)}

```


jhernand:
@Gundersanne please review.

Gundersanne (approved):
woops, sorry about that

PR #425, opened 2021-08-10
Release 0.1.199

jhernand:
Changes in this release are mainly intended to simplify packaging of the SDK in
Fedora, see issue #421 for details.

- Use `golang-jwt/jwt` instead of `dgrijalva/jwt-go`.
- Use `microcosm-cc/bluemonday` instead of `grokify/html-strip-tags-go`.
- Use `json-iterator/go` instead of `c0b/go-ordered-json`.

Related: https://github.com/openshift-online/ocm-sdk-go/issues/421

PR #427, opened 2021-08-10
dump: Skip when writing null

Gundersanne:
@jhernand might want to include this in #425


jhernand:
Included, thanks @Gundersanne.

PR #431, opened 2021-08-17
Reject impersonation

jhernand:
This patch changes the authentication handler so that it rejects tokens
that contain the `"impersonated": true` claim.

Related: https://issues.redhat.com/browse/SDA-4656


markturansky:
LGTM

PR #435, opened 2021-08-18
Update ocm-api-model version to v0.0.141

tzvatot:
@vkareh @pvasant PTAL


vkareh:
I typically separate updating the model with the new version/changelog into two separate commits, but I think this is fine

LGTM

PR #465, opened 2021-09-03
Disable jitter for retry tests

jhernand:
This patch disables jitter for retry tests because it is causing false
negatives in some situations.


jhernand:
Replaced by #466.

PR #506, opened 2021-11-30
Fix bug - service log is not terminating properly

oriAdler:
- Modify ``func (c *Connection) Close()`` to **return nil in case the connection is already closed**,
to avoid causing an error when trying to destroy OSL client when the connection
is already closed.
- Adding a test to close a closed connection returns nil.

After this PR is merged - trying to destroy a component when the connection is closed
won't cause an error.

Solve [SDA-4085](https://issues.redhat.com/browse/SDA-4085)

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -991,6 +991,11 @@ func (c *Connection) JobQueue() *jobqueue.Client {
 func (c *Connection) Close() error {
 	var err error
 
+	// in case the connection is already closed, return instead of printing an error message
+	if c.closed {
+		return nil
+	}
+
 	// Close the HTTP clients:
 	err = c.clientSelector.Close()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


oriAdler:
/assign @oriAdler 
/cc @zgalor
/help 
Regarding the test, I wasn't sure if this is the correct token ``offlineToken := MakeTokenString("Offline", 0)`` in this case.

jhernand (approved):
The token generation is OK. Looks good to me. Thanks @oriAdler.

jhernand:
@oriAdler now that this is merged, if you want to use the modified SDK in some project we will have to release it. Take a look at a previous release pull request, for example this: https://github.com/openshift-online/ocm-sdk-go/pull/501. Send a similar one and then ask someone (you can as me, for example) to review/merge/tag it.

oriAdler:
@jhernand Great, Thank you. I'll do that.

PR #512, opened 2021-12-07
Add support for status-board

djberg96:
The changes here are mainly autogenerated via the `make generate` command, and are based on https://github.com/openshift-online/ocm-api-model/pull/445.

I've also added some examples as well.

I can't speak to the formatting tweaks it seems to have made to some unrelated files, if there's a way to avoid that then please let me know.


jhernand:
This is going to fail because the tests will run the code generation tool again to verify that there are no differences. But there will be differences because you didn't update the Makefile to use the new version of the model.

To do this correctly we will need first to release a new version of the model: update the CHANGES.md file and tag it. Than change the Makefile of this project to use that new version.

djberg96:
@jhernand I bumped the Makefile version to `v0.0.154`, and tagged this branch as `v0.1.223`. If that was wrong then you will need to give me explicit instructions for what I should do exactly.

jhernand:
Before changing anything in this project we need to release the `ocm-api-model` project. I am doing that here: https://github.com/openshift-online/ocm-api-model/pull/446. Next time you can create that pull request yourself. Once that is merged this can be tested and merged.

djberg96:
@jhernand Ok, I see that you merged that, so please let me know what/when I need to do something.

BTW, I'm generally not used to adding tags to PR's, feels strange. My experience has always been that the project manager tags the release only once something is out the door.

jhernand:
Replaced by #513.

PR #544, opened 2022-01-06
Install metamodel with `go install`

jhernand:
A previous patch changed the build process so that it downloads the
released `metamodel` binary instead of building it from source.
Unfortunately that was done in a way that only works for Linux, because
it is hardcoded to download the Linux binary. It could be changed to
check the operating system and download the corresponding binary, but
that is cumbersome. Instead of that the metamodel project has been
changed so that starting with version 0.0.49 it can be installed with
`go install`, and this patch changes this project to do that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,9 +26,7 @@ model_version:=v0.0.167
 model_url:=https://github.com/openshift-online/ocm-api-model.git
 
 # Details of the metamodel to use:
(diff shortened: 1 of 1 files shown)


jhernand:
@djberg96 this should allow you to generate the code in macOS. Please check.

djberg96:
Excellent, thank you!

PR #561, opened 2022-02-03
Removes SKUs

markturansky:
Also see https://github.com/openshift-online/ocm-api-model/pull/479


markturansky:
@jhernand PTAL

jhernand (changes requested):
This is failing because it checks that there are no differences between the code generated by `make generate` and the actual code.

To make this change successful you need a new release of the model, and then update the `model_version` variable in the makefile.

markturansky:
@jhernand thanks. i don't have permissions to make a new tag in https://github.com/openshift-online/ocm-api-model/tags

markturansky:
@jhernand the contributing guide clearly states this step. thanks for gently pointing it out to me.  https://github.com/openshift-online/ocm-sdk-go/blob/main/CONTRIBUTING.md#releasing-a-new-ocm-api-model-version

Makefile
```diff
@@ -33,7 +33,7 @@ examples:
 
 .PHONY: test tests
 test tests:
-	ginkgo run -r
+	ginkgo run -r .
```
markturansky, line 36: @jhernand PTAL at this specifically.   there was a recent change here.

jhernand: I think this doesn't need to be changed. Why do you think it does?

jhernand: I will assume that you think this change is needed to make the tests pass, but it isn't. Some of the tests fail randomly because they are time sensitive and timing in the GitHub actions environment seems to be very different to our development environments. I am trying to find a solution for that.

I am closing this and replacing it with another pull request to include other changes that have been made to the model.

PR #562, opened 2022-02-04
Update to model 0.0.172

jhernand:
following:

- Remove deprecated `SKUs` endpoint.
- Remove deprecated quota summary resource and type.
- Add QuotaVersion to ClusterAuth.
- Allow adding/removing operator roles.


jhernand:
FYI: @markturansky

jhernand:
@cristianoveiga in order to make this update I had to remove the `examples/list_quota_summary.go` example. Would be nice if you can replace it with something similar that uses the current API.

cristianoveiga:
> @cristianoveiga in order to make this update I had to remove the `examples/list_quota_summary.go` example. Would be nice if you can replace it with something similar that uses the current API.

thank you very much @jhernand !

@markturansky, I understand that the quota_cost endpoint should be used from now on. Should we add a list_quota_cost.go example?

markturansky:
@cristianoveiga yes, that sounds like a good example to include if you wish.

cristianoveiga:
perfect, thanks @markturansky. I will work on that and submit a new PR .

PR #564, opened 2022-02-04
added list_quota_cost example

cristianoveiga:


cristianoveiga:
@markturansky PTAL

rawsyntax:
LGTM

examples/list_quota_cost.go
```diff
@@ -0,0 +1,76 @@
...
+	defer connection.Close()
+
+	organizationId := "" // update with your own organizationId
+
+	// Get the client for the resource that manages the collection of quota cost:
+	collection := connection.AccountsMgmt().V1().Organizations().Organization(organizationId).QuotaCost()
+
+	// Search quota cost items where quota_id starts with 'add-on':
+	response, err := collection.List().
+		Search("quota_id like 'add-on%'").
```
markturansky, line 63: Searching for a specific quoat_id is risky because they are not intended to be human (or machine) readable.  They can change.

cristianoveiga: thanks for your feedback @markturansky. Is there a particular field that you would recommend instead? I was thinking in using `type` (keeping the add-on example).

Or ... we can switch the search to something like `availability_zone` | `product`. Please advise.

markturansky: `type` is good. This is only an example showing use of the SDK.  We definitely don't want to encourage parsing the quotaID field.

cristianoveiga: sounds good. I will send another PR shortly.

PR #568, opened 2022-02-16
Release 0.1.242

zvikorn:


jhernand (changes requested):
If we are going to do the update of the model and the release in the same pull request (not what I prefer, but won't reject it) then we need to include in this pull request the changes to the `CHANGES.md` and `version.go` files.

jhernand (approved):
Looks good, thanks @zvikorn.

PR #578, opened 2022-03-04
Centralizing impersonation constants and methods.

renan-campos:
The ImpersonateUserHeader is defined and used by three different
microservices (AMS, clusters service, managed service). By defining the
constants and methods in the SDK, the three services will always be
in-sync to their definitions.

### DIFF
--- a/clustersmgmt/v1/impersonation.go
+++ b/clustersmgmt/v1/impersonation.go
@@ -1,0 +1,32 @@
+/*
+Copyright (c) 2022 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package v1 // github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1
+
+import "github.com/openshift-online/ocm-sdk-go/helpers"
+
+// ImpersonateUser is the name of the request header that is used to indicate that the caller wants
+// to impersonate another user.
+const ImpersonateUserHeader = "Impersonate-User"
+
+// Impersonation can be done by adding a "Impersonate-User" entry to the header.
+// However, use of the sdk may look cleaner by
+// adding and using the impersonation methods defined below.
+
+// Call this method to impersonate a user when adding a cluster
+func (r *ClustersAddRequest) Impersonate(username string) *ClustersAddRequest {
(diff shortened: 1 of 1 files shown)


clustersmgmt/v1/impersonation.go
```diff
@@ -0,0 +1,32 @@
...
+// ImpersonateUser is the name of the request header that is used to indicate that the caller wants
+// to impersonate another user.
+const ImpersonateUserHeader = "Impersonate-User"
+
+// Impersonation can be done by adding a "Impersonate-User" entry to the header.
+// However, use of the sdk may look cleaner by
+// adding and using the impersonation methods defined below.
+
+// Call this method to impersonate a user when adding a cluster
+func (r *ClustersAddRequest) Impersonate(username string) *ClustersAddRequest {
```
renan-campos, line 29: Users can be impersonated by calling Head(ImpersonateUserHeader, username) when making a request. But I think it may make the code more readable to have Impersonation methods for the types of requests we are making. WDYT?

jmolmo:
it seems handy .. but 
shouldn't this change be made in the model?
https://github.com/openshift-online/ocm-sdk-go/blob/main/CONTRIBUTING.md#validating-model-updates

clustersmgmt/v1/impersonation.go
jhernand: This makes sense to me, but with this approach we would have to explicitly add this method to all the methods were we want to add it. I would probably be better to add it to the code generator, so that it is automatically added to all requests. That would be here:

https://github.com/openshift-online/ocm-api-metamodel/blob/496cb39f3f7f298cf716f02e6cda9a673ed32cee/pkg/generators/golang/clients_generator.go#L798-L802

Should be something like this:

```go
func (r *{{ $requestName }}) Impersonate(user string) *{{ $requestName }} {
	helpers.AddImpersonationHeader(&r.header, user)
	return r
}
```

Then add the `AddImpersonationHeader` and the context in the generated helpers package:

https://github.com/openshift-online/ocm-api-metamodel/blob/496cb39f3f7f298cf716f02e6cda9a673ed32cee/pkg/generators/golang/helpers_generator.go#L178-L185

Should be something like this:

```go
const impersonateUserHeader = "Impersonate-User"

func AddImpersonationHeader(header *http.Header, user string) {
        AddHeader(header, impersonateUserHeader, user)
}
```

Once that is done we can release the metamodel, regenerate the SDK and we will have the new `Impersonate` method available in all the methods.

jhernand:
@jmolmo note that adding it to the model isn't possible at the moment. The model only supports query parameters, not headers.

clustersmgmt/v1/impersonation.go
jhernand: @renan-campos Another inconvenient of this approach is that the check that verifies that generated code hasn't been manually modified fails. That is because it just checks for changes in the `clustersmgmt` directory, and it detects the new file as a change, even if it isn't really a change in the generated code. If you modify the metamodel instead this will not happen.

renan-campos: @jhernand Thank you for this guidance! I created a PR with your suggested changes: https://github.com/openshift-online/ocm-api-metamodel/pull/170

PR #580, opened 2022-03-07
Release 0.1.246

jhernand:

- Update to model 0.0.178:
  - Add `managed_service` field to add-on type.
  - Add `credentials_secret` field to add-on type.
  - Add `region` field to provision shard.


jhernand:
FYI: @samira-barouti

PR #583, opened 2022-03-08
Add impersonation test

jhernand:
This patch adds a test to verify that using the `Impersonate` method
results in adding the `Impersonate-User` header.


renan-campos:
Thank you for adding this test.

PR #589, opened 2022-03-15
Update to model 0.0.182

jhernand:
following:

- Adding `service_mgmt` service.


connection.go
```diff
@@ -991,6 +992,11 @@ func (c *Connection) StatusBoard() *statusboard.Client {
 	return statusboard.NewClient(c, "/api/status-board")
 }
 
+// ServiceMgmt returns the client for the service management service.
+func (c *Connection) ServicMgmt() *servicemgmt.Client {
+	return servicemgmt.NewClient(c, "/api/service_mgmt")
+}
+
```
jhernand, line 999: @renan-campos when a new service is added most of the code is generated automatically, but this tiny method has to be written manually.

PR #612, opened 2022-03-30
Updated model to v0.0.186

samira-barouti:


jhernand:
If this is going the release change then it needs to include the update of the `version.go` file.

jhernand (approved):
Looks good, thanks @samira-barouti.

PR #615, opened 2022-03-31
Add some webrca examples

djberg96:
Some example programs I created for webrca.

### DIFF
--- a/examples/get_webrca_incident.go
+++ b/examples/get_webrca_incident.go
@@ -1,0 +1,74 @@
+/*
+Copyright (c) 2022 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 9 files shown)


djberg96:
@jhernand It doesn't look like the current test failure is related. Any ideas?

djberg96:
On a side note, when I built this locally I had to use `WebRca` instead of `WebRCA`. Not a huge deal, just thought that was strange.

jhernand (approved):
This is grea. Than you @djberg96.

jhernand:
That `WebRCA` method is the only part of the code that isn't automatically generated from the code. I wrote it manually I assumed you would want to use the normal rules for initialisms in Go: all uppercase. I guess that when you wrote this locally you decided to use `WebRca` instead. I don't have a strong opinion; if you prefer `WebRca` go ahead and submit a pull request to change it.

djberg96:
@jhernand `WebRCA` is probably expected, so I'm fine with leaving it. Thanks!

PR #616, opened 2022-04-01
Update to metamodel 0.0.53

jhernand:
following:

- Don't consider `Status` and `Error` built-in request parameters.


jhernand:
FYI: @djberg96

PR #617, opened 2022-04-01
Update to metamodel 0.0.54

jhernand:
following:

- Remove generation of experimental server code.


jhernand:
FYI: @djberg96

PR #618, opened 2022-04-01
Update to model 0.0.188

jhernand:
following:

- Add Status query param for incidents resource.


jhernand:
FYI: @djberg96

PR #632, opened 2022-04-19
Release 0.1.263

LeshemNoa:
- Update to model 0.0.195:
  - Added Import method to the HTPasswd IDP user collection.
  - Added credential request type and updated the addon type to include it

However, when I `make generate` with updated api-model, the `credential_request` resource along with its associated builder etc., and it is added to the `add_on_type`. I'm not sure why these changes were not released before as these api-model changes appear to have been made last month in this [MR](https://github.com/openshift-online/ocm-api-model/pull/530). @vkareh was this intentional? 

PTAL @jhernand


vkareh:
@LeshemNoa - that MR was opened a while ago, but it was just merged yesterday

LeshemNoa:
@vkareh so can I add it to this version release? I'll make sure to include it in the notes.

vkareh:
Yes please add all changes to the release

CHANGES.md
```diff
@@ -5,6 +5,12 @@ SDK.
 
 ## 0.1.262 Apr 14 2022
```
vkareh, line 6: This should be `## 0.1.263 Apr 19 2022`

vkareh:
@LeshemNoa You seem to be missing the `version.go` file here. Make sure you update that for the release to work.

A nitpick here is that we typically do 2 commits: one to update the api model, and another one for the actual release (CHANGES.md and version.go), they can both be part of the same PR anyway.

But whichever way we do it, you'll need to update the `version.go` file

vkareh:
LGTM

PR #642, opened 2022-05-18
Update to model 0.0.200

jhernand:
following:

- Add `hypershift.enabled` field to the cluster type.

Related: https://issues.redhat.com/browse/SDE-1586


jhernand:
FYI: @samira-barouti

PR #644, opened 2022-05-24
authentication: Allow client credential grants with basic auth

vkareh:


vkareh:
@jhernand - please take a look

vkareh:
Tested against localhost, stage, prod, both on ROSA an OCM, against RH-SSO and AWS Cognito, using various types of grants. Merging.

PR #656, opened 2022-07-12
Changed model version to latest to include delete registry credentials method

tirthct:
**Changed**
- Changed model version to latest to include delete registry credentials method


cristianoveiga:
LGTM

cristianoveiga:
@rawsyntax can you please take a look as well? 

We updated the ocm-api-model by adding the Delete method to the RegistryCredential Resource. This will be used by ocm-support-cli.

Note: you will notice changes is the _clustermgmt_ files. I reached out to the author of the changes, @renan-campos , and they confirmed that these changes are safe to be pushed to the SDK.

PR #657, opened 2022-07-13
Updated version to add delete registry credential method

tirthct:
**Changed**
- Updated version to add delete registry credential method
- Added change detail in Changes.md file


cristianoveiga:
LGTM

PR #658, opened 2022-07-22
Added capabilities property in Account_Type

tirthct:
**Changed**
- Added capabilities property in Account_Type
- Changed model version


cristianoveiga:
LGTM

PR #662, opened 2022-08-02
Bump ocm-api-model version: add update function to provision shard

tzvatot:
@zgalor @oriAdler PTAL


zgalor:
I think you are mission the `version` file

zgalor:
LGTM

PR #668, opened 2022-08-09
Draft: SDA-6553: Removed bluemonday import and added htmlsanitizer

hunterkepley:
[SDA-6553](https://issues.redhat.com/browse/SDA-6553)


hunterkepley:
No need to remove, the license for the bluemonday project [BSD-3] is fine for our use case

PR #679, opened 2022-09-16
Add OSD Fleet Management client service in SDK

gabriel-farache:
W hen bumping to model 0.0.216, the OSD FM model was added but the OSD FM client service in `connection.go` was forgotten.
This PR fixes this.

### DIFF
--- a/version.go
+++ b/version.go
@@ -18,4 +18,4 @@ limitations under the License.
 
 package sdk
 
-const Version = "0.1.287"
+const Version = "0.1.288"
--- a/version.go
+++ b/version.go
@@ -18,4 +18,4 @@ limitations under the License.
 
 package sdk
 
-const Version = "0.1.288"
+const Version = "0.1.287"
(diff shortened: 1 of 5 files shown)


pkliczewski:
@pvasant please take a look

gabriel-farache:
@etabak if you can take a look, that would be great

version.go
```diff
@@ -18,4 +18,4 @@ limitations under the License.
 
 package sdk
 
-const Version = "0.1.287"
+const Version = "0.1.288"
```
pvasant, line 21: we update version only when we tag so we can remove it  from this PR

gabriel-farache: @pvasant What do you mean by that?
I will reset the version to 0.1.287 then

tzvatot:
LGTM

andreadecorte:
we should also add osdfleetmgmt on https://github.com/openshift-online/ocm-sdk-go/blob/main/.github/workflows/check-pull-request.yaml#L91 @gabriel-farache 
otherwise LGTM

andreadecorte:
LGTM

gabriel-farache:
@pvasant @tzvatot can any one of you merge this PR? Thank you :)

tzvatot:
LGTM

PR #696, opened 2022-12-06
V0.1.300

ciaranRoche:
Update to model 0.0.240 and Release SDK 0.1.300


ciaranRoche:
[test]

jhernand:
@ciaranRoche looks like you formatted the code in this patch with a tool that removes some empty comment lines, and replaces consecutive single quotes with a double quote. That is why the "Generate" check is failing. You need to make sure you use exactly the same version used by that check:

https://github.com/openshift-online/ocm-sdk-go/blob/ca6adc712288a44d1362cfd9e5614abf721b7faa/.github/workflows/check-pull-request.yaml#L80

Either that or update the project to use the version of "goimports" that you used, as it produces nicer code.

lucasponce (approved):

PR #697, opened 2022-12-08
Bump pr check 1.19

ciaranRoche:


ciaranRoche:
@jhernand can you PTAL at this, do you think it best to go with this, or should I just roll back my versions locally and we stick with 1.18?

jhernand (approved):
This looks completely reasonable to me.

PR #761, opened 2023-04-20
updated model to v0.0.277

cristianoveiga:


Makefile
```diff
@@ -18,7 +18,7 @@
 export CGO_ENABLED=0
 
 # Details of the model to use:
-model_version:=v0.0.276
+model_version:=v0.0.277
```
lucasponce, line 21: I think you could remove the version here.

@ciaranRoche we typically separate the changes and let to bump the version when we are going to release, right?

ciaranRoche: No, this looks right, its only a model bump, you need to bump this var to run `make generate` to get the changes in.

cristianoveiga: Thanks for taking a look @lucasponce , @lucasponce . 

When this one merged, I will create another MR with create the release.

cristianoveiga: It looks like someone else already updated the model (the release v0.0.278 is already available).

I will close this PR.

PR #766, opened 2023-04-27
Release v0.1.336

paradigm:
Update model to v0.0.280, adding account_group_id to role_binding


rawsyntax:
Typically the update is done in a separate pull request and the release is done by itself. For example:

https://github.com/openshift-online/ocm-sdk-go/pull/745
https://github.com/openshift-online/ocm-sdk-go/pull/746

accountsmgmt/v1/access_token_auth_builder.go
```diff
@@ -20,6 +20,8 @@ limitations under the License.
 package v1 // github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1
 
 // AccessTokenAuthBuilder contains the data and logic needed to build 'access_token_auth' objects.
+//
+//
```
rawsyntax, line 24: I'm not sure what is generating these changes, but it looks like too many files are touched. We'll need to figure out how to fix this before merging.

@gdbranco please take a look

gdbranco: Need >= go@1.19 higher and >= goimports@0.4.0 if I'm not mistaken

`make fmt` might also resolve it, but with those versions it should work with make generate
@vkareh do you remember the command for fmt?

vkareh: I think just `make fmt` should clean them up. I get those changes every time until I remember to format it

renan-campos:
Closing this PR, since we're a few releases ahead of this by this point.

PR #798, opened 2023-06-28
Add govcloud parameter to cloud region 

eldar101:
Adding the govcloud parameter, in order to include it in region posting testing for OCM-backend-tests.


eldar101:
@etabak @oriAdler Please review and let me know If I'm missing something.
Thanks!

tzvatot:
@eldar101 this should be added to ocm-api-model first, and then bumped to here. Please see CONTRIBUTING.md.

igoihman (changes requested):
This change can't be merged without a corresponding design proposal and agreement about this API change

PR #811, opened 2023-07-20
Added Marketplace GCP billing model to accounts and clusters mgmt

tirthct:
**Changed**
- Added Marketplace GCP billing model to accounts and clusters mgmt
- Covered changes from https://github.com/openshift-online/ocm-api-model/commit/d8ed56b84be194881d8faf0838f137ed1de94802


clustersmgmt/v1/cloud_providers_client.go
```diff
@@ -100,6 +101,14 @@ func (r *CloudProvidersListRequest) Impersonate(user string) *CloudProvidersList
 	return r
 }
 
+// FetchRegions sets the value of the 'fetch_regions' parameter.
```
cben, line 104: FYI the param is actually called `fetchRegions`, correctly sent on L180, just the comments generated by metamodel are still inaccurate when overriding url param names with `@http` override.

(These CS APIs are not important for AMS anyway)

Lir10:
shouldn't we update the ocm-api-model version in `version.go` ?

PR #837, opened 2023-09-11
OCM-3277 | SDK Bump to use model v0.0.317

den-rgb:


robpblake:
@den-rgb I think we can close this one now?

PR #842, opened 2023-09-20
feat: bump ocm-api-model to v0.0.319

machi1990:
See https://github.com/openshift-online/ocm-api-model/releases/tag/v0.0.319 for changes that are included in this new model version

/cc @tzvatot


machi1990:
We'll be no longer taking this direction and I'll work on reverting the changes from the API model. 
I'll close this JIRA.

PR #844, opened 2023-09-20
OCM-2437 | feat: Added support for /api/clusters_mgmt/v1/aws_inquiries/sts_account_roles to the SDK

robpblake:
This PR adds support for `/api/clusters_mgmt/v1/aws_inquiries/sts_account_roles` to the SDK. 

The backend is already implemented in CS, this exposes the functionality via the SDK.


robpblake:
This is picking up the additional changes from #842, so ideally we need that merged first and then I can rebase.

clustersmgmt/v1/environment_type_json.go
```diff
@@ -93,26 +102,30 @@ func readEnvironment(iterator *jsoniter.Iterator) *Environment {
 			break
 		}
 		switch field {
+		case "backplane_url":
```
robpblake, line 105: Changes from #842, which should be merged first.

oriAdler:
@robpblake we need to update the `CHANGES.md` file as well.
Similar PR https://github.com/openshift-online/ocm-sdk-go/pull/786/files#diff-d975bf659606195d2165918f93e1cf680ef68ea3c9cab994f033705fea8238b2

oriAdler:
We also need to update the `version.go` file.

clustersmgmt/v1/environment_type_json.go
robpblake: These changes have been removed from this PR as they are no longer required.

robpblake:
@oriAdler Updated `version.go` and `CHANGES.md` as requested. Thanks for the pointer 👍

oriAdler (approved):

PR #855, opened 2023-10-23
OCM-4313: Update golang in ocm-sdk-go v1.21

machi1990:


machi1990:
PTAL @zgalor @tzvatot @ciaranRoche once approved, I'd like to get https://github.com/openshift-online/ocm-sdk-go/pull/856 in as well and a new releases be created

miguelsorianod:
The changes lgtm. Just a side question: Have you run `go mod tidy` after updating to Go 1.20?
If you did I am guessing there were some changes related to Go's management of dependencies, as I see several of them being added as indirect

miguelsorianod:
> The changes lgtm. Just a side question: Have you run `go mod tidy` after updating to Go 1.20? If you did I am guessing there were some changes related to Go's management of dependencies between Go 1.16 and 1.20, as I see several of them being added as indirect

I am guessing it must have been in Go 1.17: https://tip.golang.org/doc/go1.17

machi1990:
> The changes lgtm. Just a side question: Have you run `go mod tidy` after updating to Go 1.20? If you did I am guessing there were some changes related to Go's management of dependencies between Go 1.16 and 1.20, as I see several of them being added as indirect

Yes, I've run `go mod tidy` and almost all of the changes in the go.mod have been due to that.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/openshift-online/ocm-sdk-go
 
-go 1.16
+go 1.20
```
zgalor, line 3: why not 1.21?

.github/workflows/check-pull-request.yaml
```diff
@@ -29,9 +29,8 @@ jobs:
     strategy:
       matrix:
         go:
-        - 1.18
-        - 1.19
-        - '1.20'
+        - "1.20"
+        - "1.21"
```
zgalor, line 33: why was this added?

machi1990: I added 1.20 & 1.21 here since I seem to remember that there are projects on 1.20 & 1.21 e.g  CS & AMS respecitvely, so I thought we could test against both

zgalor: and why not bump the sdk to 1.21?

tzvatot: I guess that will force any dependent project to upgrade too. Which is not too bad. It's mainly CS, AMS, OSL, AS, MSS, FM.

machi1990: Okay, let's give it a `go` to 1.21. 
At worst, dependent will be forced to update. Otherwise the go compiler will try to build the dependent package while ensuring the compatibleness promise.

machi1990: Updated

machi1990: > It's mainly CS, AMS, OSL, AS, MSS, FM.

@tzvatot @zgalor Just for information - there are other services using the ocm-sdk-go module e.g RHACS https://github.com/stackrox/acs-fleet-manager/blob/main/go.mod#L36

tzvatot: When dependent components would like to upgrade to latest OCM SDK, then they will need to update go latest Go as well.

I think that's reasonable. Can you please also run it with BU?

machi1990: I've tagged BU on the JIRA.

machi1990:
@zgalor @tzvatot PTAL

tzvatot (approved):

PR #862, opened 2023-11-01
OCM-4601 | chore: bump api-model to v0.0.332

gdbranco:
Related issue: OCM-4601


ciaranRoche (approved):
/lgtm

CHANGES.md
```diff
@@ -3,6 +3,11 @@
 This document describes the relevant changes between releases of the OCM API
 SDK.
 
+## 0.1.380
+- Update model version v0.0.332
+  - Add `AdditionalInfraSecurityGroupIds` to `AWS` type
+  - Add `AdditionalControlPlaneSecurityGroupIds` to `AWS` type
```
cristianoveiga, line 9: I'm seeing changes related to https://github.com/openshift-online/ocm-api-model/pull/854, should we add them here?

ciaranRoche: This ^^^ we have got out of sync between the version on the api model also. We need a new tag in the api-model and we also need to update the changes here

gdbranco: @ciaranRoche 332 is the latest, but I didn't notice that 331 hadn't been created in time, so I created only the tag later. Over here it is just missing to include the changes of the bump

gdbranco: https://github.com/openshift-online/ocm-api-model/releases

Fixed

PR #876, opened 2023-11-30
OCM-3227: Add support for regionalized urls

tylercreller:
* Adds Regionalized URL support to ConnectionBuilder

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -241,6 +241,9 @@ func (b *ConnectionBuilder) Region(region string) *ConnectionBuilder {
 	if b.err != nil {
 		return b
 	}
+	if region == "" {
+		return b.AlternativeURL("", DefaultURL)
+	}
 	return b.AlternativeURL("", fmt.Sprintf(FormattedDefaultRegionURL, region))
 }
 
--- a/connection.go
+++ b/connection.go
@@ -51,12 +51,12 @@ import (
 // Default values:
 const (
 	// #nosec G101
-	DefaultTokenURL           = authentication.DefaultTokenURL
-	DefaultClientID           = authentication.DefaultClientID
-	DefaultClientSecret       = authentication.DefaultClientSecret
-	DefaultURL                = "https://api.openshift.com"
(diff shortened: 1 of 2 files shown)


macgregor (changes requested):

connection.go
```diff
@@ -232,6 +233,17 @@ func (b *ConnectionBuilder) URL(url string) *ConnectionBuilder {
...
 
+// Region builds the base URL of an API gateway based on an OCM region name. The default is
+// `https://api.%s.openshift.com`. Where `%s` is replaced by the region name.
+//
+// This method should be used in-lieu of URL when connecting to a regionalized API gateway.
+func (b *ConnectionBuilder) Region(region string) *ConnectionBuilder {
+	if b.err != nil {
+		return b
+	}
+	return b.AlternativeURL("", fmt.Sprintf(FormattedDefaultRegionURL, region))
```
markturansky, line 244: this would return `https://api..opeshift.com` for commercial prod (where empty region id is meaningful)

connection.go
```diff
@@ -566,6 +578,7 @@ func (b *ConnectionBuilder) Load(source interface{}) *ConnectionBuilder {
 		URL              *string           `yaml:"url"`
 		AlternativeURLs  map[string]string `yaml:"alternative_urls"`
 		TokenURL         *string           `yaml:"token_url"`
+		Region           *string           `yaml:"region"`
```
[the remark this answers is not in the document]
markturansky: it's a valid point.  because region is now meaningful, then `--region=integration` works exactly as you think. no region is commercial prod.

markturansky (changes requested):

connection.go
tylercreller: The SDKs default URL is prod. Not attaching region or URL defaults to prod. No need to support commercial prod with the Region attribute.

tylercreller: https://github.com/openshift-online/ocm-sdk-go/blob/main/connection.go#L133

tylercreller: updated to short-circuit when a value is not provided

PR #882, opened 2023-12-18
OCM-5281 | Feat | Add regions support from ocm shards

tirthct:
**Changed**
- Add regions support from ocm shards

**Tested**
-Regions should return correct value for endpoints : tested


rh_region.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+	return regions, nil
+}
+
+func GetRhRegion(ocmServiceUrl string, regionName string) (Region, error) {
+	regions, err := GetRhRegions(ocmServiceUrl)
+	if err != nil {
+		return Region{}, err
+	}
+	regionName = fmt.Sprintf("rh-%s", regionName)
```
macgregor, line 40: the "rh-" prefix assumption is still here and in OCM-5281, again, dont force a key-prefix naming convention, just look for the `regionName` in the map as it is passed to the function. 

As its currently implemented, this function would find "singapore" in the map `{ "rh-singapre": foo }` but it would not find "rh-singapore" (because it would look for "rh-rh-singapre") which is confusing. We also dont want to dynamically add the "rh-" prefix because then neither "singapore" nor "rh-singapore" would work for the map  `{ "singapre": foo }` because now the map data is "wrong". 

Smart keys are rarely worth it and should only be used if you have a very hard key format requirement that you need to enforce.

tirthct: Changed to logic to check for `contains` of a region (we can change this as and when the shards json is finalized)

rh_region.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func DetermineRegionDiscoveryUrl(ocmServiceUrl string) (string, error) {
+	baseUrl, err := url.Parse(ocmServiceUrl)
+	if err != nil {
+		return "", err
+	}
+	regionDiscoveryHost := "api.openshift.com"
+	//TODO: Remove the OR condition from this if statement before the MR merge
+	if strings.HasSuffix(baseUrl.Hostname(), "integration.openshift.com") || true {
```
macgregor, line 55: get rid of the //TODO comment and the `|| true` and i think this MR will be good to go

tirthct: Removed TODO comment and always true condition

macgregor:
lgtm, thanks @tirthct

PR #885, opened 2024-01-09
OCM-4961 | Feat | Add authentication using OAuth2 and PCKE

tirthct:
**Changed**
- Add auth.go that uses OAuth2 and PCKE for getting access token
- Add example to verify authentication

**Tested**
- Auth should redirect to Red Hat SSO and get access token from there : tested


examples/verify_authentication.go
```diff
@@ -0,0 +1,54 @@
...
+	// Create the connection, and remember to close it:
+	token, err := authentication.VerifyLogin("cloud-services", "")
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Can't get token: %v\n", err)
+		os.Exit(1)
+	}
+	connection, err := sdk.NewConnectionBuilder().
+		Logger(logger).
+		Tokens(token).
+		URL("http://localhost:8000").
```
tirthct, line 37: TODO: remove this once we get a permission for prod SSO to localhost redirection

authentication/auth.go
```diff
@@ -0,0 +1,82 @@
...
+
+func VerifyLogin(clientID string, clientSecret string) (string, error) {
+	authToken = ""
+	ctx = context.Background()
+	conf = &oauth2.Config{
+		ClientID:     clientID,
+		ClientSecret: clientSecret,
+		Scopes:       []string{"openid"},
+		Endpoint: oauth2.Endpoint{
+			AuthURL:  "https://sso.stage.redhat.com/auth/realms/redhat-external/protocol/openid-connect/auth",
```
tirthct, line 51: TODO: replace with prod SSO when localhost redirection is available

authentication/auth.go
```diff
@@ -0,0 +1,82 @@
...
+			TokenURL: "https://sso.stage.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token",
```
tirthct, line 52: TODO: replace with prod SSO when localhost redirection is available

authentication/auth.go
```diff
@@ -0,0 +1,86 @@
...
+		done <- 0
+	}
+}
+
+func VerifyLogin(clientID string, clientSecret string) (string, error) {
+	authToken = ""
+	ctx = context.Background()
+	conf = &oauth2.Config{
+		ClientID:     clientID,
+		ClientSecret: clientSecret,
```
tylercreller, line 51: I don't believe we'll need to provide Client Secret for this flow. Can you confirm with Josh?

authentication/auth.go
```diff
@@ -0,0 +1,86 @@
...
+	time.Sleep(1 * time.Second)
+	for {
+		if authToken != "" {
+			return authToken, nil
+		}
+		if time.Now().After(twoMinTimer) {
+			return authToken, fmt.Errorf("Time expired")
+		}
+	}
+}
```
tylercreller, line 86: Let's add some more comments to this file. Just want to make sure the flow is as easy to understand as possible. It seems like there are a few nuances here with magic numbers and workflow.

authentication/auth.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	authToken = tok.AccessToken
+	_, err = io.WriteString(w, "Login successful! Please close this window and return back to CLI")
+	if err != nil {
+		log.Fatal(err)
+	}
+}
+
+func serve(done chan int) {
+	log.Fatal(http.ListenAndServe(":9998", nil))
```
tylercreller, line 40: I'm not entirely sure what is happening here but it looks like we are relying on log.Fatal to close the connection? If so, let's try and properly close this without log.fatal

tirthct: httpListenAndServe returns an error, if one occurs, and log.Fatat is responsible for logging the error and closing one. I don't this we have any other approach for closing the connection as the documentation only specifies this one (Ref: https://pkg.go.dev/net/http#ListenAndServe)

tylercreller: There exists a shutdown command https://pkg.go.dev/net/http#Server.Shutdown

tylercreller: Example usage with some similar patterns https://stackoverflow.com/questions/39320025/how-to-stop-http-listenandserve

authentication/auth.go
tirthct: Removed client secret from input params

authentication/auth.go
tirthct: Added

authentication/auth.go
```diff
@@ -0,0 +1,99 @@
...
+
+	// Exchange will do the handshake to retrieve the initial access token.
+	tok, err := conf.Exchange(ctx, code, oauth2.VerifierOption(verifier))
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Get the access token and ask user to go back to CLI
+	authToken = tok.AccessToken
+	_, err = io.WriteString(w, "Login successful! Please close this window and return back to CLI")
```
tylercreller, line 37: Are we able to automatically close the window after X seconds?

```
Login successful! Your browser will close in X seconds...

```

authentication/auth.go
tirthct: Used http.Server to host and shutdown the handler

tylercreller:
LGTM once prod SSO is ready to go

examples/verify_authentication.go
tirthct: Removed

authentication/auth.go
tirthct: Replaced

authentication/auth.go
tirthct: Replaced

authentication/auth.go
```diff
@@ -73,8 +73,8 @@ func VerifyLogin(clientID string) (string, error) {
 		ClientSecret: "",
 		Scopes:       []string{"openid"},
 		Endpoint: oauth2.Endpoint{
-			AuthURL:  "https://sso.stage.redhat.com/auth/realms/redhat-external/protocol/openid-connect/auth",
-			TokenURL: "https://sso.stage.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token",
+			AuthURL:  "https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/auth",
+			TokenURL: "https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token",
```
tylercreller, line 77: Let's use the SSO constant from `authentication/transport_wrapper.go`. And create new constants for Redirect URL and Port (shared with the serve function)

tylercreller (approved):

PR #891, opened 2024-01-25
OCM-5759: Add device auth

tylercreller:
* Adds device auth capabilities
* Updates entry point func name for auth code to match device auth
* OAuth functions now return refresh tokens instead of access tokens

### DIFF
--- a/examples/device_auth.go
+++ b/examples/device_auth.go
@@ -28,7 +28,19 @@ func main() {
 	}
 
 	// Create the connection, and remember to close it:
-	token, err := authentication.InitiateDeviceAuth(clientId)
+	cfg := &authentication.DeviceAuthConfig{
+		ClientID: clientId,
+	}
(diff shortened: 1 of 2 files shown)


examples/device_auth.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+	defer connection.Close()
+
+	collection := connection.AccountsMgmt().V1().CurrentAccount()
+
+	// Retrieve current account information
+	_, err = collection.Get().
+		SendContext(ctx)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Can't retrieve capabilities: %s\n", err)
```
tirthct, line 53: Nit : Can't retrieve current account information

tirthct:
LGTM

examples/device_auth.go
tylercreller: Good catch, updated

tirthct (approved):

PR #895, opened 2024-01-26
Allow Connection to avoid using the default authentication transport

miguelsorianod:
# Description

This MR implements being able to use a `Connection` object without authentication being performed on the OCM SDK Go client side.

The motivation for this is to be able to connect to an OCM service that does not require authentication using the OCM SDK Go. For example, when running Cluster Service locally without authentication enabled.

A solution is proposed and described in this PR. I also listed alternatives that I could think of, each with its own tradeoffs and implications. I think in this case it is probably a case of desired tradeoffs.

I also added new example example as part of the PR contents.

## Implemented solution

Users of the OCM SDK Go can create a Connection by using a ConnectionBuilder. Before this PR users could create a ConnectionBuilder by using the `NewConnectionBuilder` constructor method.

This PR adds a new UnauthenticatedConnectionBuilder constructor method that returns a ConnectionBuilder that when is built returns a Connection that does not perform authentication (this is, that it does not make use of the authentication.TransportWrapper http transport).

With this, users of the OCM SDK Go have two ways of creating a ConnectionBuilder:
* NewConnectionBuilder
* NewUnauthenticatedConnectionBuilder

A new `bool includeDefaultAuthenticationTransportWrapper` unexported attribute has been included in the Connection type. When it is set to false then the default authentication wrapper attribute is not set in the Connection object which leaves it as nil. Methods of the Connection type that make use of the default authentication wrapper attribute it check if is nil and if it is they return "noops" like empty strings, nil errors, ... . The includeDefaultAuthenticationTransportWrapper attribute is set to false when NewUnauthenticatedConnectionBuilder is used

### Implications of the new solution

* The Connection type has authentication-related methods: TokenURL, Client, User, Scopes, Tokens, TokensContext. Because no authentication is used when using NewUnauthenticatedConnectionBuilder, these methods have to return something anyway. "noop" returns have been implemented in that case (when authnWrapper is nil) like nil strings and nil errors, no matter what the input parameters of those methods are.
* Now we have two constructor methods for ConnectionBuilder
* Now `authnWrapper` can be nil in the Connection type when no authentication is used for it
* Due to the previous point, there are now nil checks for the authnWrapper attribute in different methods in the Connection type that make use of it. If in the future the authnWrapper is used in other places nil checks should be implemented for it too. An alternative would be assuming the user won't call methods that need authentication. However if that were to happen nil exceptions would return
* When calling NewUnauthenticatedConnectionBuilder, if the user calls authentication-specific builder methods those will basicall have no effect

## Alternative solutions

* Instead of having 2 different constructor methods for ConnectionBuilder, update the existing one to receive as a parameter whether the default authentication wrapper is to be used for the connection or not, or receive a ConnectionBuilder options struct with one of its attributes being whether the default authentication wrapper is to be used for the connection or not.
  * This solution is a "breaking change" in the sense that code (users or within ocm sdk go itself) that makes use of NewConnectionBuilder needs to be updated when using the code after this PR (in a new ocm sdk go version)
* Change the behavior of  the `TransportWrapper` method in the ConnectionBuilder to be a set instead of an append
  * This solution is a breaking change due to the `TransportWrapper` method changing its behavior
  * This solution also makes responsible the user to be aware and set among the default transport wrappers
* Add a new method in the ConnectionBuilder type that is something like bool UseDefaultAuthenticationWrapper which is
  set to true by default
  * This solution has the same implications as the proposed solution replacing having two constructor methods for a new additional method on the builder

Aside from the previous alternative solutions, there are other possible approaches that imply a bigger redesign. If we analyze the Connection type, one of the characteristics is that it assumes that it always uses authentication. This is reflected by it offering methods that are authentication-specific only: TokenURL, Client, User, Scopes, Tokens, TokensContext.
This other approaches involve trying to decouple authentication from the Connection type.
Some of this approaches are:
* Approach 1
  * Change Connection to not have default authentication transport at all. No authentication related attributes nor methods
  * Create a new AuthenticatedConnection type that contains a Connection and basically adds the authentication to it
  * This approach is a breaking change
  * This approach implies that now there are two different connection types
  * This approach might also imply that now there two different connection builder types
  * If a part of the code in the SDK uses Connection type then it won't be able to use authentication specific methods
* Approach 2
  * Create an UnauthenticatedConnection type. It has no authentication related attributes nor methods
  * Refactor the Connection type to contain an UnauthenticatedConnection instance. This type has all the methods UnauthenticatedConnection has plus authentication-specific related methods
  * This approach implies that now there are two different connection types. For example, ConnectionWithNoRetry
  * Another note is that if we want to add/remove other behaviors we might end up with an explosion of number of types.
  * This approach might imply you also need multiple different connection builders
  * OCM SDK Go code itself
  * If a part of the code in the SDK uses UnauthenticatedConnection type then it won't be able to use authentication specific methods
* Approach 3
  * Create a Connection interface that does not include any authentication specific behavior nor retry behavior
  * Create a BaseConnection struct type that implements the Connection interface
  * Create an AuthenticatedConnection struct type that implements the Connection interface plus authentication specific methods. This type contains a Connection instance as part of it
  * Create a RetriableConnection struct type that includes a Connection interface. It implements also the Connection interface and adds additional retry-specific methods too
  * This approach is a breaking change
  * This approach also implies that if the Connection is passed around in the code and for some reason we need for example authentication specific data then we need to type assert it to an AuthenticatedConnection and deal appropriately if it is not the expected type
  * This approach also leaves the responsability of instantiating the desired connection behavior. This might also imply multiple connection builders at different levels.

## Pending
- [x] Testing. For now I have not implemented any potential tests. I will evaluate it once an agreement to a solution has been reached.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -86,6 +86,8 @@ type ConnectionBuilder struct {
 	retryJitter       float64
 	transportWrappers []func(http.RoundTripper) http.RoundTripper
 
+	includeDefaultAuthnTransportWrapper bool
+
 	// Metrics:
 	metricsSubsystem  string
 	metricsRegisterer prometheus.Registerer
@@ -132,11 +134,20 @@ func NewConnectionBuilder() *ConnectionBuilder {
 		urlTable: map[string]string{
 			"": DefaultURL,
 		},
-		retryLimit:        retry.DefaultLimit,
-		retryInterval:     retry.DefaultInterval,
-		retryJitter:       retry.DefaultJitter,
-		metricsRegisterer: prometheus.DefaultRegisterer,
+		retryLimit:                          retry.DefaultLimit,
+		retryInterval:                       retry.DefaultInterval,
+		retryJitter:                         retry.DefaultJitter,
+		metricsRegisterer:                   prometheus.DefaultRegisterer,
+		includeDefaultAuthnTransportWrapper: true,
 	}
+}
+
+// NewConnectionBuilder creates a Builder that knows how to create connections
+// without authentication
+func NewUnauthenticatedConnectionBuilder() *ConnectionBuilder {
+	connectionBuilder := NewConnectionBuilder()
+	connectionBuilder.includeDefaultAuthnTransportWrapper = false
+	return connectionBuilder
 }
 
 // Logger sets the logger that will be used by the connection. By default it uses the Go `log`
@@ -739,24 +750,34 @@ func (b *ConnectionBuilder) BuildContext(ctx context.Context) (connection *Conne
 		loggingWrapper = wrapper.Wrap
 	}
 
-	// Create the authentication wrapper:
-	authnWrapper, err := authentication.NewTransportWrapper().
+	// Initialize the client selector builder:
+	clientSelectorBuilder := internal.NewClientSelector().
 		Logger(b.logger).
-		TokenURL(b.tokenURL).
-		User(b.user, b.password).
-		Client(b.clientID, b.clientSecret).
-		Tokens(b.tokens...).
-		Scopes(b.scopes...).
 		TrustedCAs(b.trustedCAs...).
-		Insecure(b.insecure).
-		TransportWrapper(metricsWrapper).
-		TransportWrapper(loggingWrapper).
-		TransportWrappers(b.transportWrappers...).
-		MetricsSubsystem(b.metricsSubsystem).
-		MetricsRegisterer(b.metricsRegisterer).
-		Build(ctx)
-	if err != nil {
-		return
+		Insecure(b.insecure)
+
+	var authnWrapper *authentication.TransportWrapper
+	if b.includeDefaultAuthnTransportWrapper {
+		// Create the authentication wrapper:
+		authnWrapper, err = authentication.NewTransportWrapper().
+			Logger(b.logger).
+			TokenURL(b.tokenURL).
+			User(b.user, b.password).
+			Client(b.clientID, b.clientSecret).
+			Tokens(b.tokens...).
+			Scopes(b.scopes...).
+			TrustedCAs(b.trustedCAs...).
+			Insecure(b.insecure).
+			TransportWrapper(metricsWrapper).
+			TransportWrapper(loggingWrapper).
+			TransportWrappers(b.transportWrappers...).
+			MetricsSubsystem(b.metricsSubsystem).
+			MetricsRegisterer(b.metricsRegisterer).
+			Build(ctx)
+		if err != nil {
+			return
+		}
+		clientSelectorBuilder.TransportWrapper(authnWrapper.Wrap)
 	}
 
 	// Create the retry wrapper:
@@ -771,11 +792,7 @@ func (b *ConnectionBuilder) BuildContext(ctx context.Context) (connection *Conne
 	}
 
 	// Create the client selector:
-	clientSelector, err := internal.NewClientSelector().
-		Logger(b.logger).
-		TrustedCAs(b.trustedCAs...).
-		Insecure(b.insecure).
-		TransportWrapper(authnWrapper.Wrap).
+	clientSelector, err := clientSelectorBuilder.
 		TransportWrapper(metricsWrapper).
 		TransportWrapper(retryWrapper.Wrap).
 		TransportWrapper(loggingWrapper).
@@ -873,25 +890,39 @@ func (c *Connection) Logger() logging.Logger {
 }
 
 // TokenURL returns the URL that the connection is using request OpenID access tokens.
+// An empty string is returned if the connection does not use authentication.
 func (c *Connection) TokenURL() string {
+	if c.authnWrapper == nil {
+		return ""
+	}
 	return c.authnWrapper.TokenURL()
 }
 
 // Client returns OpenID client identifier and secret that the connection is using to request OpenID
 // access tokens.
+// Empty strings are returned if the connection does not use authentication.
 func (c *Connection) Client() (id, secret string) {
-	id, secret = c.authnWrapper.Client()
+	if c.authnWrapper != nil {
+		id, secret = c.authnWrapper.Client()
+	}
 	return
 }
 
 // User returns the user name and password that the is using to request OpenID access tokens.
+// Empty strings are returned if the connection does not use authentication.
 func (c *Connection) User() (user, password string) {
-	user, password = c.authnWrapper.User()
+	if c.authnWrapper != nil {
+		user, password = c.authnWrapper.User()
+	}
 	return
 }
 
 // Scopes returns the OpenID scopes that the connection is using to request OpenID access tokens.
+// An empty slice is returned if the connection does not use authentication.
 func (c *Connection) Scopes() []string {
+	if c.authnWrapper == nil {
+		return []string{}
+	}
 	return c.authnWrapper.Scopes()
 }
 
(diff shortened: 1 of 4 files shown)


miguelsorianod:
cc @zgalor @etabak for your thoughts on this. At the moment I opened the PR as a draft so we can iterate and agree on the solution. Thanks.

connection.go
```diff
@@ -136,9 +138,18 @@ func NewConnectionBuilder() *ConnectionBuilder {
 		retryInterval:     retry.DefaultInterval,
 		retryJitter:       retry.DefaultJitter,
 		metricsRegisterer: prometheus.DefaultRegisterer,
+		includeDefaultAuthenticationTransportWrapper: true,
```
zgalor, line 141: np: indentation

tzvatot:
Did you considered the approach of exposing `AuthTransportWrapper()` function?

This way you can provide your own "no-auth" transport, or otherwise, the builder will use the default.

The advantage of that is that it's following the builder pattern, where with the current design you are introducing different building functions.

That said, I don't have a strong objection to take the current approach, just raising the option above.

connection.go
```diff
@@ -873,25 +890,41 @@ func (c *Connection) Logger() logging.Logger {
...
+		return ""
+	}
 	return c.authnWrapper.TokenURL()
 }
 
 // Client returns OpenID client identifier and secret that the connection is using to request OpenID
 // access tokens.
+// Empty strings are returned if the connection does not use authentication.
 func (c *Connection) Client() (id, secret string) {
+	if c.authnWrapper == nil {
```
zgalor, line 905: np:
```
if c.authnWrapper != nil {
  id, secret = c.authnWrapper.Client()
}
return
```

connection.go
```diff
@@ -873,25 +890,41 @@ func (c *Connection) Logger() logging.Logger {
...
 	id, secret = c.authnWrapper.Client()
 	return
 }
 
 // User returns the user name and password that the is using to request OpenID access tokens.
+// Empty strings are returned if the connection does not use authentication.
 func (c *Connection) User() (user, password string) {
+	if c.authnWrapper == nil {
+		return "", ""
+	}
```
zgalor, line 917: same as previous comment

token.go
```diff
@@ -48,8 +51,14 @@ func (c *Connection) Tokens(expiresIn ...time.Duration) (access, refresh string,
 // expired, this method will do it and will return an error if it fails.
 //
 // If new tokens are needed the request will be retried with an exponential backoff.
+// The returned access and refresh tokens are empty strings if the
+// connection does not use authentication. In that case no error is
+// returned either
 func (c *Connection) TokensContext(ctx context.Context, expiresIn ...time.Duration) (access,
 	refresh string, err error) {
+	if c.authnWrapper == nil {
```
zgalor, line 59: same as the comments above

zgalor:
> Did you considered the approach of exposing `AuthTransportWrapper()` function?
> 
> This way you can provide your own "no-auth" transport, or otherwise, the builder will use the default.
> 
> The advantage of that is that it's following the builder pattern, where with the current design you are introducing different building functions.
> 
> That said, I don't have a strong objection to take the current approach, just raising the option above.

The problem with that approach is that usage is that it's too implicit.
I rather users create their own specific auth or noAuth connection and not change if for each Build().

That said, if the demand for different auth arises, we might need to adopt such approach

connection.go
miguelsorianod: updated

miguelsorianod:
> > > Did you considered the approach of exposing `AuthTransportWrapper()` function?
> > > This way you can provide your own "no-auth" transport, or otherwise, the builder will use the default.
> > > The advantage of that is that it's following the builder pattern, where with the current design you are introducing different building functions.
> > > That said, I don't have a strong objection to take the current approach, just raising the option above.
> > 
> > 
> > The problem with that approach is that usage is that it's too implicit. I rather users create their own specific auth or noAuth connection and not change if for each Build().
> > That said, if the demand for different auth arises, we might need to adopt such approach
> 
> Well, the user can also build a connection without credentials, try to use it, and it will fail. That's the builder pattern - it does not guarantee that you will get an object that will work, it just gives you easy options to build one. It seems like with this approach we are mixing the factory pattern with the builder pattern.

I guess this solution is similar to the alternative solution I described `Add a new method in the ConnectionBuilder type that is something like bool UseDefaultAuthenticationWrapper which is set to true by default`, allowing you to set a custom authn wrapper instead of enable/disable the default one.

The major implication of this proposed approach is that it would probably require a redesign of `authentication.TransportWrapper`:  It is currently not possible to pass a NoAuthn instead because `authentication.TransportWrapper` is not an interface but a struct. Aside from that, there is also `authentication.TransportWrapperBuilder`. We should decide what to do regarding to that. Should we pass a NoAuthnWrapper type directly or somehow update the `authentication.TransportWrapperBuilder` to support getting a NoAuthnWrapper? there are some assumptions in those types where authentication is inherently within them (see all the attributes they contain)

miguelsorianod:
I added a test that uses the unauthenticated connection builder (as described in the description of the PR)

connection.go
```diff
@@ -132,13 +134,22 @@ func NewConnectionBuilder() *ConnectionBuilder {
...
+		retryInterval:                       retry.DefaultInterval,
+		retryJitter:                         retry.DefaultJitter,
+		metricsRegisterer:                   prometheus.DefaultRegisterer,
+		includeDefaultAuthnTransportWrapper: true,
 	}
 }
 
+// NewConnectionBuilder creates a Builder that knows how to create connections
+// without authentication
+func NewUnauthenticatedConnectionBuilder() *ConnectionBuilder {
```
miguelsorianod, line 147: An alternative name could be NewNoAuthConnectionBuilder() or NewNoAuthnConnectionBuilder() or NewNoAuthenticationConnectionBuilder().

Let me know if you have any preference instead of the currently set name

tzvatot:
> bool UseDefaultAuthenticationWrapper which is set to true by default

The bool option is not the same as providing the auth wrapper itself. 
It's the difference between given the builder instructions on how to build (which is essentially a factor), than to provide it with the actual building blocks to build with.

I don't have a string objection of using the current approach, it has it's advantages. All I'm saying that it's a bit of anomaly  when using a builder.

miguelsorianod:
> > bool UseDefaultAuthenticationWrapper which is set to true by default
> 
> The bool option is not the same as providing the auth wrapper itself. It's the difference between giving the builder instructions on how to build (which is essentially a factory), than to provide it with the actual building blocks to build with.

Yes. I said that it was a "similar" approach, not the same. Similar as in it is a setting in the builder, but instead of enable/disable it would set the desired authentication wrapper itself

miguelsorianod:
As we discussed,

It is not clear whether in the near future we might need to add more flexibility around the authentication area. It is quite possible we might end up needing to do more changes, redesign or even remove the current approach in the PR.

The usage of a connection without authentication should be limited to internal usage only so we are fine with removing or redesigning in an incompatible way the changes performed in this PR.

I think we can go ahead with the merging of the PR if it looks good to you. If you think the names used should be different let me know too.

connection.go
tzvatot: This works for me

tzvatot (approved):
LGTM

connection.go
zgalor: I'm fine either way

PR #898, opened 2024-01-31
OCM-2780 | fix: Don't error when response is 204 and no content-type

hunterkepley:
Also fixes: https://issues.redhat.com/browse/OCM-4087

Original ticket: https://issues.redhat.com/browse/OCM-2780

This MR makes it so that when we have a 204 no content, we return empty response regardless of a missing content-type in the header. This was causing errors leading to nil responses in the SDK within the Keycloak FedRAMP environment

### DIFF
--- a/send.go
+++ b/send.go
@@ -99,6 +99,9 @@ func (c *Connection) RoundTrip(request *http.Request) (response *http.Response, 
 	}
 
 	// Check that the response content type is JSON:
+	if response.StatusCode == http.StatusNoContent {
+		return
+	}
 	err = internal.CheckContentType(response)
 	if err != nil {
 		return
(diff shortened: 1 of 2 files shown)


send.go
```diff
@@ -99,6 +99,9 @@ func (c *Connection) RoundTrip(request *http.Request) (response *http.Response,
 	}
 
 	// Check that the response content type is JSON:
+	if response.StatusCode == http.StatusNoContent {
+		return
+	}
```
tylercreller, line 104: IMO this would be better suited in the CheckContentType function itself. It is also used in the transport wrapper and would prevent any future headache if it is used any where else in the future.

hunterkepley: Moved

tylercreller (approved):

ciaranRoche (approved):

PR #905, opened 2024-02-13
OCM-4965: Secure store tweaks

tylercreller:
* Add test for invalid and empty keyrings
* Additional Error Standardization
* Add support that allows the caller to define a keyring to target
* Add error handling for denied Keychain access due to permissions

### DIFF
--- a/authentication/securestore/main.go
+++ b/authentication/securestore/main.go
@@ -5,12 +5,12 @@ import (
 	"compress/gzip"
 	"fmt"
 	"io"
+	"strings"
 
 	"github.com/99designs/keyring"
 )
 
 const (
-	SecureStoreConfigKey = "securestore"       // OCM_CONFIG key to enable secure OS store
 	KindInternetPassword = "Internet password" // MacOS Keychain item kind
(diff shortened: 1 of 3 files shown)


tirthct (approved):

authentication/securestore/main.go
```diff
@@ -103,32 +118,42 @@ func UpsertConfigToKeyring(creds []byte) error {
...
 
 	err = ring.Remove(ItemKey)
+	if err != nil {
+		if err == keyring.ErrKeyNotFound {
+			// Ignore not found errors, key is already removed
+			return nil
+		}
+
+		if strings.Contains(err.Error(), "Keychain Error. (-25244)") {
+			return fmt.Errorf("%s\nThis application may not have permission to delete from the Keychain. Please check the permissions in the Keychain and try again", err.Error())
```
macgregor, line 139: Friendly reminder that when wrapping errors your should use "%w" rather than %s or %v to properly preserve the original error in case callers want it: https://gosamples.dev/wrap-unwrap-errors/

PR #911, opened 2024-02-26
Migrate keychain to use non-CGO libraries

tylercreller:
* Refactors keychain keyring storage to use a non-CGO library
* Adds tests for `keychain`, `wincred`, `pass`
* Modifies PR check CI to have separate runs per platform for keyring tests


authentication/securestore/main.go
```diff
@@ -64,11 +64,14 @@ func IsBackendAvailable(backend string) (isAvailable bool) {
...
 
 // AvailableBackends provides a slice of all available backend keys on the current OS.
-//
-// Note: CGO_ENABLED=1 is required for darwin builds (enables OSX Keychain)
 func AvailableBackends() []string {
 	b := []string{}
 
+	if runtime.GOOS == "darwin" {
+		// Assume Keychain is always available on Darwin. It will not return from keyring.AvailableBackends()
+		b = append(b, "keychain")
```
macgregor, line 72: is this because keychain is built into the OS? How far back does that go in Mac history? just making sure we dont need to do a more sophisticated check and thow some kind of "not supported" exception on older macs. Or check explicitly the app/version installed...

Its not a huge concern, just trying to error-proof if its simple enough to do.

macgregor:
for context, why do we want to use a non-CGO library?

authentication/securestore/main.go
tylercreller: Yep exactly. [First release was 1999](https://en.wikipedia.org/wiki/Keychain_(software)). 

Unofficially, if for some reason the version of macOS does not contain keychain, errors will be handled during CRUD operations.

Officially, our CLIs officially should only support latest OS releases and this should be a non-issue.

tylercreller:
> for context, why do we want to use a non-CGO library?

CGO breaks cross-compilation. Requiring CGO for our CLIs requires workarounds for cross-compilation or building darwin binaries on dedicated macOS CI runners. For ROSA this is a large amount of overhead due to their current release processes.

authentication/securestore/main.go
```diff
@@ -207,6 +210,55 @@ func ValidateBackend(backend string) error {
...
+	return creds, nil
+}
+
+func keychainUpsert(creds []byte) error {
+	compressed, err := compressConfig(creds)
+	if err != nil {
+		return err
+	}
+
+	err = gokeyring.Set(ItemKey, ItemKey, string(compressed))
```
cristianoveiga, line 239: According to the docs, the second argument here is the user:

```
// Set password in keyring for user.
func Set(service, user, password string) error
```

I wonder if this will cause issues in case there's more than one user

authentication/securestore/main.go
```diff
@@ -207,6 +210,55 @@ func ValidateBackend(backend string) error {
 	return nil
 }
 
+func keychainGet() ([]byte, error) {
+	credentials, err := gokeyring.Get(ItemKey, ItemKey)
+	if err != nil && err != gokeyring.ErrNotFound {
+		return []byte(credentials), err
+	} else if err == gokeyring.ErrNotFound {
+		return []byte(""), nil
+	}
```
cristianoveiga, line 219: The linter is suggesting using err.Is due to `Comparison with errors using equality operators fails on wrapped errors`

authentication/securestore/main.go
```diff
@@ -116,13 +121,15 @@ func UpsertConfigToKeyring(backend string, creds []byte) error {
...
 
 // RemoveConfigFromKeyring will remove the credentials from the first priority OS secure store.
-//
-// Note: CGO_ENABLED=1 is required for OSX Keychain and darwin builds
 func RemoveConfigFromKeyring(backend string) error {
 	if err := ValidateBackend(backend); err != nil {
 		return err
 	}
 
+	if runtime.GOOS == "darwin" && backend == "keychain" {
```
cristianoveiga, line 129: nit: this check was used a couple of times, it probably deserves its own function

authentication/handler_test.go
```diff
@@ -1,3 +1,6 @@
+//go:build linux
+// +build linux
```
cristianoveiga, line 2: is this really required? Where would it fail (windows or mac)?

authentication/securestore/main.go
tylercreller: We are setting user as the same Item Key to maintain parity with the way configuration is handled today. If a user logs in with a different account, it will just overwrite what is in the keyring with the new info.

tylercreller: image

authentication/handler_test.go
tylercreller: Looks like it is happy, resolving this

cristianoveiga:
LGTM

tirthct (approved):

PR #916, opened 2024-02-28
Bump to api-model to v0.0.358

machi1990:
This adds the DomainPrefix field onto the clusters API


machi1990:
The linting issue is also in `main`. Is this a known issue?

machi1990:
> The linting issue is also in `main`. Is this a known issue?

The issue is in main as well. I don't think it should be a blocker for this

machi1990:
Not needed, we agreed to with https://github.com/openshift-online/ocm-sdk-go/pull/918

PR #950, opened 2024-05-02
OCM-7308 | feat: Generated a new version of the OCM-SDK to support KubeletConfigs on HCP MachinePools

robpblake:
This PR adds support for KubeletConfig on HCP MachinePools. This will allow customers to set a custom PIDs limit on their HCP MachinePools.


clustersmgmt/v1/kubelet_config_builder.go
```diff
@@ -60,10 +61,17 @@ func (b *KubeletConfigBuilder) Empty() bool {
...
+	b.name = value
+	b.bitmap_ |= 8
+	return b
+}
+
 // PodPidsLimit sets the value of the 'pod_pids_limit' attribute to the given value.
 func (b *KubeletConfigBuilder) PodPidsLimit(value int) *KubeletConfigBuilder {
 	b.podPidsLimit = value
-	b.bitmap_ |= 8
+	b.bitmap_ |= 16
```
den-rgb, line 74: is the bit being changed here to support a higher number of pids on hcp clusters?

robpblake: @den-rgb I think it's because there is now the `name` field on the `KubeletConfig` in addition to the `podPidsLimit` field. All of this code is generated, so not something I'm specifically controlling.

den-rgb: I see okay :+1:

ciaranRoche: FWIW the bitmap is keeping track of what fields are being set, in this case the 5th bit as we have added name which now takes the 4th bit. Lovely binary notation

ciaranRoche (approved):
/lgtm

PR #958, opened 2024-05-28
OCM-5952: Server name inference for regionalized OCM redirects

tylercreller:
- Allow go tls to infer the `ServerName` from host address to enable redirects with TLS


macgregor:
lgtm, thanks for driving this home

rawsyntax (approved):

PR #984, opened 2024-08-06
chore: bump ocm-api-model to v0.0.388

hunterkepley:


Makefile
```diff
@@ -28,6 +28,7 @@ export CGO_ENABLED=0
 
 # Details of the model to use:
 model_version:=v0.0.387
```
marcolan018, line 30: this line should be removed

marcolan018, line 30: this line should be removed

gdbranco (approved):

PR #989, opened 2024-08-27
OCM-10678 | fix: tag must have value with map[string]interface{}

aaraj7:


clustersmgmt/v1/aws_node_pool_builder.go
```diff
@@ -33,7 +33,7 @@ type AWSNodePoolBuilder struct {
 	instanceType               string
 	rootVolume                 *AWSVolumeBuilder
 	subnetOutposts             map[string]string
-	tags                       map[string]string
+	tags                       map[string]interface{}
```
gdbranco, line 36: If this requires a change it should be in the api model instead, the files for the SDK are auto generated
https://github.com/openshift-online/ocm-sdk-go/blob/c1abafd8c1d342f7782fa983b69e9aaa01491d35/clustersmgmt/v1/aws_node_pool_builder.go#L17

Tried to follow the ticket, but couldn't find the reason why this is a requirement

aaraj7: Sorry, I automated the test case through old sdk , not needed this PR now, Closing it

PR #992, opened 2024-09-03
OCM-10871: Expose hcp_namespace field for Hypershift Model

devppratik:
This PR updates the model to expose `hcp_namespace` field in the Hypershift model which is not available in SDK but can be queried via the API using the `ocm get cluster $CLUSTER_ID/hypershift `endpoint

Related PR for model - https://github.com/openshift-online/ocm-api-model/pull/980


andreadecorte:
not needed, changes autogenerated.

PR #1016, opened 2024-12-15
Add ARO HCP v1alpha1 root resource

nimrodshn:
## What this does

-  Include the necessary changes for derived structs using the `@ref` statement.
- Add the ARO HCP v1alpha1 root resource as well as the Clusters derived struct.


ahitacat:
LGTM :tada:

zgalor (approved):
lgtm

PR #1018, opened 2025-01-20
OCM-12069 | (feat): Add cluster migrations with examples

AlexVulaj:
Adds cluster_migration resources with examples.

This PR depends on https://github.com/openshift-online/ocm-api-model/pull/1013, and will require a change to `model_version` in the Makefile once that PR is merged.


CONTRIBUTING.md
```diff
@@ -53,7 +53,7 @@ included with the version update. The following is an example update to version
 - Add `config_managed` attribute to the `RoleBinding` type.
 ```
 
-Submit an MR with the CHANGES.md modification and review/merge.
+Submit an MR with the CHANGES.md and version.go modifications and review/merge.
```
ciaranRoche, line 56: There is no version.go file in the ocm-api-model. Its only needed in the SDK. So we can remove these changes from the MR

AlexVulaj: My brain thought this was under the ocm-sdk changes - good catch. Fixed.

PR #1033, opened 2025-03-09
ARO-15553 | Update model and meta model

oriAdler:
Add the new inflight check endpoints.


oriAdler:
@zgalor @nimrodshn PTAL

nimrodshn:
Very nice ! :partying_face:

PR #1035, opened 2025-03-20
New release v0.1.462

ahitacat:
## 0.1.462
- Updated model version to v0.0.414
  - Added node pool endpoint resources for aro_hcp.
  - Added node pool state model in clusters_mgmt.
  - Added node pool status resources for aro_hcp.
- Updated meta model version to v0.0.67
  - Added async methods

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,14 @@
 This document describes the relevant changes between releases of the OCM API
 SDK.
 
+## 0.1.462
+- Updated model version to v0.0.414
+  - Added node pool endpoint resources for aro_hcp
+  - Added node pool state model in clusters_mgmt
+  - Added node pool status resources for aro_hcp
+- Updated meta model version to v0.0.67
+  - Added async methods
+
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -3,6 +3,14 @@
 This document describes the relevant changes between releases of the OCM API
 SDK.
 
+## 0.1.462
+- Updated model version to v0.0.414
+  - Added node pool endpoint resources for aro_hcp.
```
miguelsorianod, line 8: Nitpick: n this same PR in some places we end a list element with the `.` character and not in some others. We should be consistent.

For the latest few releases we haven't been using `.` so I propose the same here.

As a side note, taking a look at the file itself in general we are not being consistent about that unfortunately

ahitacat: Oh, good catch. Fix it.

miguelsorianod (approved):

miguelsorianod:
lgtm

PR #1048, opened 2025-06-11
leadership flag enhancement: precheck handler added

rcampos2029:
Enables the configuration of a precheck function that, when set, will run before the leadership flag is checked. The precheck function returns two values: a boolean indicating whether the leadership check should occur, and an error indicating if an error occurred when running the precheck. When an error occurs during precheck, the error is logged and the leadership check is skipped.


ciaranRoche (approved):
/lgtm

PR #1055, opened 2025-06-25
bump the ocm-api model to dbec927aa153

cristianoveiga:


arohcp/v1alpha1/clusters_client.go
```diff
@@ -50,15 +50,15 @@ func NewClustersClient(transport http.RoundTripper, path string) *ClustersClient
...
+// Add creates a request for the 'async_add' method.
 //
 // Provision a new cluster and add it to the collection of clusters.
 //
 // See the `register_cluster` method for adding an existing cluster.
 func (c *ClustersClient) Add() *ClustersAddRequest {
 	return &ClustersAddRequest{
 		transport: c.transport,
-		path:      c.path,
+		path:      path.Join(c.path, "async_add"),
```
vkareh, line 61: This seems like a potentially breaking change... Where is it coming from?

vkareh: Ah! I see it's part of the ARO API.

PR #1056, opened 2025-06-30
Add a script to simplify current update process

nimrodshn:
cc @miguelsorianod @etabak @deads2k @ahitacat


hack/update-ocm-api-model.sh
deads2k, line 1: blocking comment: make a `hack/verify-foo` script that ensures all the SHAs for ocm-api-model are the same.

non-blocking suggestion: take an optional arg to specify a level.

non-blocking suggestion: create a make target to run this script.

hack/update-ocm-api-model.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
```
tzvatot, line 1: `#!/bin/bash -e` please

tzvatot (approved):

PR #1057, opened 2025-07-01
Update the release process of the ocm-api-model in the CONTRIBUTING file

nimrodshn:


CONTRIBUTING.md
```diff
@@ -39,44 +39,24 @@ Hard tabs are required rather than spaces.
...
 make update
 ```
 
-In most cases, the ocm-api-model version will be incremented prior to generation. To increment the ocm-api-model
-version, update the `model_version` constant in [Makefile](Makefile).
+The `./hack/update-model.sh` script will ensure the `ocm-api-model` modules are all up to date with the latest commit SHA across the OCM-SDK project.
+To verify that they are all in-sync one can use the `./hack/verify-model-version.sh` script.
 
 Whenever an update is made, ensure that the corresponding example in [examples](examples) is also updated where
 necessary. Any new endpoints should have a new example created.
```
machi1990, line 62:
```suggestion
.. Any new endpoints should have a new example created.
```

is this a requirement or a recommendation? @tzvatot @nimrodshn It's more like a requirement but I don't see it being followed, perhaps we might want to consider rephrasing this.

CONTRIBUTING.md
```diff
@@ -39,44 +39,24 @@ Hard tabs are required rather than spaces.
 ## Releasing a new OCM API Model version
```
machi1990, line 39: The section title is talking about "release" but the description underneath( the last part) is talking more about updating/bumping the model in the SDK.

I wonder how applicable is this section to the ocm-sdk-go repo. I find that it has its place in the ocm-api-model.

PR #1058, opened 2025-07-02
Add an optional commit SHA argument to the update-model script

nimrodshn:
cc @deads2k @etabak @machi1990

### DIFF
--- a/hack/update-model.sh
+++ b/hack/update-model.sh
@@ -1,6 +1,34 @@
 #!/bin/bash -e
 
-# This script ensures that all the api-model submodules used across the project are updated to the latest verison.
+# ==============================================================================
+# Script: update-model.sh
+#
+# This script ensures that all the `ocm-api-model` submodules used across the
+# project are updated to the latest version or pinned to a specific commit SHA.
+#
+# The OCM SDK can be generated simply by running the following after all changes have been made:
+#
+# ./hack/update-model.sh
+# make update
+#
+# USAGE:
+#   ./update-model.sh [COMMIT_SHA]
+#
+# ARGUMENTS:
+#   COMMIT_SHA (optional) - If provided, all listed modules will be pinned to
+#                           this specific commit instead of updating to latest.
+#
+# EXAMPLES:
+#   ./update-model.sh
+#     → Updates all modules to their latest versions.
+#
+#   ./update-model.sh f67fb59980981bdc81d95d1379a82da5bcec57bf
+#     → Pins all modules to the specified commit SHA.
+#
+# ==============================================================================
+
+# Optional commit SHA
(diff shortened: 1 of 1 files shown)


hack/update-model.sh
```diff
@@ -1,6 +1,29 @@
...
 
-# This script ensures that all the api-model submodules used across the project are updated to the latest verison.
+# ==============================================================================
+# Script: update-model.sh
+#
+# This script ensures that all the `ocm-api-model` submodules used across the
+# project are updated to the latest version or pinned to a specific commit SHA.
+#
+# USAGE:
+#   ./update-model.sh [COMMIT_SHA]
```
machi1990, line 10: can we add a line on the doc about this https://github.com/openshift-online/ocm-sdk-go/blob/979171be01871fe6b1aaf4cca8409f30422e1d8d/CONTRIBUTING.md#updating-the-ocm-sdk ?

hack/update-model.sh
```diff
@@ -16,6 +39,9 @@ MODULES=(
 )
 
 echo "Updating Go modules..."
+if [ -n "$COMMIT_SHA" ]; then
+  echo "Using specific commit SHA: $COMMIT_SHA"
```
tzvatot, line 43: @nimrodshn can you add a sanity check that the commit sha is legit? i.e. it actually refers to an existing git sha1

tzvatot (changes requested):

hack/update-model.sh
```diff
@@ -1,6 +1,47 @@
...
+REPO_URL="https://github.com/openshift-online/ocm-api-model.git"
+
+if [ -n "$COMMIT_SHA" ]; then
+  echo "Checking if commit SHA exists in remote repo..."
+  if git ls-remote "$REPO_URL" | grep -q "^$COMMIT_SHA"; then
+    echo "✅ Commit $COMMIT_SHA found in remote repository."
+  else
+    echo "❌ Commit SHA $COMMIT_SHA not found in remote $REPO_URL"
+    exit 1
+  fi
```
machi1990, line 43: Why not just let `go get ` to it's thing?
e.g getting a non existing `hjdhjsd898` commit 
```
go get github.com/openshift-online/ocm-api-model/model@hjdhjsd898                                                      [
go: github.com/openshift-online/ocm-api-model/model@hjdhjsd898: invalid version: unknown revision hjdhjsd898
```

tzvatot: I don't mind how to validate (git log or go get) - I was just looking to make sure if a user gives a wrong SHA it will fail. As long as that happens - I'm happy :)

nimrodshn: Ok. I am removing this validation than.

PR #1064, opened 2025-07-17
Revert "chore: update ocm api metamodel to v0.0.422"

miguelsorianod:
This reverts commit c99d8cb6a9083e4bf517bd46292f1aacd1b2211c.

We found that the metamodel generator is not generating code correctly after the attempt to fix the limtation of number of attributes per resource.

We revert the commit that updated the SDK to leverage the new metamodel generator until we fix the issues with the metamodel generator and we generate new code in the model.


tzvatot:
I hope we won't need this as latest version of the SDK is ready

PR #1067, opened 2025-07-21
Add git hook to validate a pushed tag matches the Version constant

tzvatot:
@ciaranRoche PTAL

### DIFF
--- a/hooks/pre-push
+++ b/hooks/pre-push
@@ -1,0 +1,135 @@
+#!/bin/bash
+
+#
+# Git pre-push hook to validate that tag versions match the Version constant in version.go
+#
+# This hook is called with the following parameters:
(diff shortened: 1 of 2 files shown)


hooks/pre-push
```diff
@@ -0,0 +1,135 @@
...
+    echo "WARNING: $1" >&2
+}
+
+print_info() {
+    echo "INFO: $1" >&2
+}
+
+# Function to extract version from version.go
+get_version_from_file() {
+    if [[ ! -f "version.go" ]]; then
```
ciaranRoche, line 31: This func assumes we are in the root of the directory :thinking:  should we validate that to either include a better error message, or some traversal to get to the root

ciaranRoche (approved):
LGTM

PR #1071, opened 2025-07-24
Use ocm-api-model v0.0.426

nimrodshn:


nimrodshn:
v0.0.428 was a version I had pushed while experimenting with 1077 :face_exhaling: which Chitanya accidentally used while bumping the sdk - I had since deleted it so we need to downgrade back to the latest version v0.0.426.

tzvatot (approved):
/lgtm

PR #1080, opened 2025-08-26
OCM-17788 | feat: image mirrors for rosa hcp

davidleerh:


clustersmgmt/v1/image_mirror_client.go
```diff
@@ -0,0 +1,580 @@
...
+	status int
+	header http.Header
+	err    *errors.Error
+}
+
+// Status returns the response status code.
+func (r *ImageMirrorDeleteResponse) Status() int {
+	if r == nil {
+		return 0
+	}
```
ziccardi, line 289: Are there any cases where calling `Status()` on a `nil` `ImageMirrorDeleteResponse` instance is actually valid or intended?
If not, I think we should remove this check to avoid silently hiding real errors.

davidleerh: Thank you. I see what you mean.

I believe this is generated from the ocm-api-metamodel as below:
https://github.com/openshift-online/ocm-api-metamodel/blob/main/pkg/generators/golang/clients_generator.go#L695-L700

and all the client files generated carry this issue.

I'll go ahead and start a sep. thread for it in our channels. Thank you.

For now, I will go ahead and merge this change request as the issue mentioned is not specific to this change but rather due to hour the ocm-api-metamodel generates the response classes in client files.

ziccardi: @tzvatot I see your point. If this has been done by design and we accept the risk, I'm ok with that.

> @ziccardi the above code snippet example is showing incorrect use of the SDK.

That was just an example to show what I was talking about. If we introduce some bug somewhere and we erroneously return or pass a `nil` response no-one will notice it since no error is returned when invoking `Status` or `Error` on the `nil` object.

> because any code using nil receiver will break if we change this behaviour

So, does this mean there are situations where returning a `nil` response and then calling methods like `Status` or `Error` on that `nil` object is actually correct? If invoking those methods on a `nil` response is supposed to be a bug, changing the behavior would simply help expose those bugs rather than hide them.

> Example on how k8s does it;
https://github.com/kubernetes/client-go/blob/0341f077c9d600b7a68d6f637cbd5da9ace4df3d/rest/config.go#L199

The Kubernetes example is a different scenario: that method returns a string representation of the config object, and the "<< nil >>" string makes sense as a textual output of a `nil` config.
In our case, what’s returned isn’t related to a `nil` response. Also, within the same object, our handling isn’t consistent: some methods check for a nil receiver (like `Status`, `Error`), while others (such as `Get` and `{{ $getterName }}`) do not:
https://github.com/openshift-online/ocm-api-metamodel/blob/24204320348aab1202ffe2a6ddb77c879e711db4/pkg/generators/golang/clients_generator.go#L728C34-L728C51

I see that at this point it might not be feasible to change this. Let’s just stay alert to the kinds of bugs this approach can potentially mask.

tzvatot: > So, does this mean there are situations where returning a nil response and then calling methods like Status or Error on that nil object is actually correct? If invoking those methods on a nil response is supposed to be a bug, changing the behaviour would simply help expose those bugs rather than hide them.

You are correct it exposes a bug, and I cannot think of any flow where using a nil receiver is a correct use of the SDK. But given what I saw, there are tests written by developers that are misusing the SDK with a nil receiver (I recently fixed some). If we change this behaviour, and those projects bump, they will fail on their CI. 

Since we cannot tell what code was written widely around the entire OCM ecosystem, my concern in changing this behaviour is that it will create a lot of disruption for everyone. It make sense if this change exposes client side bugs. But that is a difficult argument to make if someone wrote 100 unit tests doing code like the example you showed, and now need to go and fix all of them.

PR #1089, opened 2025-10-03
OCM-18794 | feat: add OCM object response helpers for testing framework

gdbranco:
Add RespondWithOCMObject and RespondWithOCMObjectMarshal helper functions to simplify testing scenarios involving OCM SDK objects. These functions handle JSON marshaling and HTTP response setup, reducing boilerplate code in test suites.


testing/servers_test.go
```diff
@@ -0,0 +1,152 @@
...
+		// Create a mock cluster object (we'll use a simple struct for testing)
+		cluster, err := v1.NewCluster().
+			ID("test-cluster").
+			Name("Test Cluster").
+			Build()
+		Expect(err).ToNot(HaveOccurred())
+
+		// Configure the server to respond with the cluster
+		server.AppendHandlers(
+			RespondWithOCMObjectMarshal(200, cluster, v1.MarshalCluster),
```
tzvatot, line 39: NP: use the http.StatusOK const instead of hard coded number 200

testing/servers_test.go
```diff
@@ -0,0 +1,152 @@
...
+		)
+
+		// Make a request
+		resp, err := http.Get(server.URL())
+		Expect(err).ToNot(HaveOccurred())
+		defer resp.Body.Close()
```
tzvatot, line 45: Need to use ginko's `After` as any assertion failure will cause skipping this.

testing/servers_test.go
```diff
@@ -0,0 +1,152 @@
...
+		cluster := &v1.Cluster{}
+		server.AppendHandlers(
+			RespondWithOCMObjectMarshal(200, cluster, "not-a-function"),
+		)
+
+		resp, err := http.Get(server.URL())
+		Expect(err).ToNot(HaveOccurred())
+		defer resp.Body.Close()
+
+		Expect(resp.StatusCode).To(Equal(500))
```
tzvatot, line 80: Same comment - use http.InternalServerError const

testing/servers_test.go
```diff
@@ -0,0 +1,152 @@
...
+	})
+})
+
+// TestRespondWithOCMObjectMarshal provides unit tests for the marshal function
+func TestRespondWithOCMObjectMarshal(t *testing.T) {
```
tzvatot, line 85: Some of the test use ginkgo while this in is using native go test. What's the motivation for using both?

testing/servers_test.go
```diff
@@ -0,0 +1,152 @@
...
+		ID("test-cluster-123").
+		Name("Test Cluster for Unit Test").
+		Build()
+	if err != nil {
+		t.Fatalf("Failed to create test cluster: %v", err)
+	}
+
+	// Create a test request
+	req := httptest.NewRequest("GET", "/test", nil)
+	rr := httptest.NewRecorder()
```
tzvatot, line 97: Would like to avoid single or 2 char variable - those are hard to read and follow.

tzvatot:
AI assisted? if so please add the label

tzvatot (changes requested):

testing/servers_test.go
gdbranco: AI generated, had not notice on the early review

testing/servers_test.go
gdbranco: Did you mean DeferCleanup? Changed to that just in case, using after would be more cumbersome due to setting it up as external variables and relying on closing later

tzvatot: Yes, my apologies, I meant `DeferCleanup`.

tzvatot (approved):

